PR #11, opened 2023-02-01
Switch to static linking go code for targets that support it

neonphog:
RESOLVES https://github.com/holochain/tx5/issues/40

Restarting this effort in light of the requirement for ios.


crates/tx5-go-pion-sys/build.rs
```diff
@@ -184,25 +184,87 @@ fn go_build() {
...
+    let manifest_path = std::env::var("CARGO_MANIFEST_DIR")
+        .map(std::path::PathBuf::from)
+        .expect("error reading manifest dir");
+
+    let cp = |f: &'static str| {
+        let mut a = manifest_path.clone();
+        a.push(f);
+        let mut b = out_dir.clone();
+        b.push(f);
+        std::fs::copy(a, b).expect("failed to copy go file");
```
ThetaSinner, line 203: Technically not only used for Go files, it's a very small thing but something like `Go project file` might be a bit clearer if this were to fail for some reason?

crates/tx5-go-pion-sys/build.rs
```diff
@@ -184,25 +184,87 @@ fn go_build() {
...
+    cp("peerconnection.go");
+    cp("go.sum");
+    cp("go.mod");
+
+    match TARGET.link_type {
+        LinkType::Dynamic => {
+            let mut lib_path = out_dir.clone();
+
+            match TARGET.go_os {
+                Some("darwin") | Some("ios") => {
```
ThetaSinner, line 219: Used to reference `"macos" | "ios" | "tvos"` where we now have translations to go_os: `macos` -> `darwin` and `ios` -> `ios`. Should `tvos` flow through here?

Also, do we need to support tvos? I can't think of a use-case for tx5 on a TV but I guess we may as well be consistent about building on the platforms we can :)

crates/tx5-go-pion-sys/build.rs
```diff
@@ -56,122 +133,45 @@ fn go_unzip_vendor() {
...
-        }
-        "openbsd" => {
-            cmd.env("GOOS", "openbsd");
-        }
-        "netbsd" => {
-            cmd.env("GOOS", "netbsd");
-        }
-        _ => (),
+    if let Some(go_os) = TARGET.go_os {
+        cmd.env("GOOS", go_os);
```
ThetaSinner, line 151: GOOS should always be set right? I wonder if neither the go_os or the go_arch should be optional and instead of the `None` in the `default` impl above this should be a build error for platforms we don't recognise?

ThetaSinner (approved):
To my level of knowledge, this looks good. I do have a few small suggestions which are mostly in the form of questions because this isn't stuff I'm that familiar with :)

My only top level comment would be about regular verification. Is it worth including dynamic/static builds as part of the CI on different platforms and at least assert that a file is produced even if there isn't a straight forward way to check that they work?

crates/tx5-go-pion-sys/build.rs
```diff
@@ -1,6 +1,83 @@
...
+            "android" => Some("android"),
+            "dragonfly" => Some("dragonfly"),
+            "freebsd" => Some("freebsd"),
+            "openbsd" => Some("openbsd"),
+            "netbsd" => Some("netbsd"),
+            _ => None,
+        };
+
+        let mut link_type = match target_os.as_str() {
+            "windows" | "android" => LinkType::Dynamic,
```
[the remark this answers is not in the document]
neonphog: Yeah, alas, golang doesn't even support static linking on windows or android, which is why they must be dynamic. Static would be preferable everywhere otherwise.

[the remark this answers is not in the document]
neonphog: Technically, it supports it on windows, it just isn't compatible with rust.

crates/tx5-go-pion-sys/src/lib.rs
```diff
@@ -13,6 +13,27 @@
...
+    any(target_os = "macos", target_os = "ios", target_os = "tvos"),
+    link(name = "Security", kind = "framework")
+)]
+#[cfg_attr(
+    any(target_os = "macos", target_os = "ios", target_os = "tvos"),
+    link(name = "resolv")
+)]
+extern "C" {}
+
+#[cfg(all(link_dynamic, link_static))]
```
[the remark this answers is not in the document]
neonphog: It is possible to manually specify `RUSTFLAGS="--cfg link_*" cargo`, so I felt like it was worth checking at this level too.

crates/tx5-go-pion-sys/build.rs
neonphog: These are now based on the [valid targets that go can build for](https://github.com/golang/go/blob/master/src/go/build/syslist.go#L14-L33) since that's all we can support anyways. They don't seem to differentiate tvos, I'm not sure if that's just included in ios or what.

crates/tx5-go-pion-sys/build.rs
neonphog: Technically GOOS only needs to be set if we are cross-compiling. Otherwise it will just default to whatever it detects the host system to be. But yes, for our purposes, I think we want tighter control over this. If someone needs to build for a target that errors this list, they can open an issue and we can sort it out then.

neonphog: Done.

crates/tx5-go-pion-sys/build.rs
```diff
@@ -1,6 +1,83 @@
...
+            link_type,
+        }
+    }
+}
+
+static TARGET: once_cell::sync::Lazy<Target> =
+    once_cell::sync::Lazy::new(Target::default);
+
 fn main() {
+    println!("cargo:warning={:?}", *TARGET);
```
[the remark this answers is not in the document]
ThetaSinner: Interesting, well if that's just a limitation of the build scripts then that's fine. I can see why that's useful for debugging

ThetaSinner:
That makes sense, so the default paths are already being tested. That's probably enough for now. It would be good to cover the non-default cases at some point but maybe that can wait until somebody is using them for something?

ThetaSinner:
This looks good to me!

crates/tx5-go-pion-sys/build.rs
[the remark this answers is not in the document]
neonphog: Went ahead and commented them out so we don't worry folks building holochain.

PR #21, opened 2023-03-27
signal server versioning

neonphog:
- Set up the signal server to support multiple protocol versions (specified in the path)
- Set up the client to specify a single protocol version in the path when connecting
- Client connection urls will *not* contain the version, because that will be applied by the particular client that is connecting
- Adds a test that proves we get an error when a bad version is specified


peeech (approved):
Logic looks great. I wasn't able to run tests locally due to missing Go, but actions are passing :-)

PR #25, opened 2023-04-07
Ability to ban remote ids in endpoint (in lieu of connection close)

neonphog:
Since tx5 endpoints work as a message queue, opening / retrying connections as needed under the hood, there isn't a direct analog for "Closing" a connection. Yet, holochain still needs this ability if we don't want to speak to a remote node.

This PR introduces the concept of a remote id "ban" where we won't send messages to, accept messages from, or allow connections to a particular remote id for a specified time length. We can consider / configure the length of this ban period in the holochain integration PR, but perhaps ~10 seconds would function as a starting point, since, if they try to re-establish a connection after that time period, we will simply ban them again.

### DIFF
--- a/crates/tx5/src/endpoint.rs
+++ b/crates/tx5/src/endpoint.rs
@@ -148,6 +148,13 @@ impl Ep {
         self.state.listener_sig(sig_url)
     }
 
+    /// Close down all connections to, fail all outgoing messages to,
+    /// and drop all incoming messages from, the given remote id,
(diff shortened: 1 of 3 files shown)


neonphog:
holochain integration preview: https://github.com/holochain/holochain/pull/2207

thedavidmeister (changes requested):
seems like we should use the time window blocks rather than invent a separate ban system, if we think that is a perf issue then let's optimise it?

PR #32, opened 2023-05-08
harden lib loading against race conditions

neonphog:


crates/tx5-core/src/file_check.rs
```diff
@@ -0,0 +1,194 @@
...
+                    // in the worst fallback, write another tmp file and use it
+                    write(file_data)?
+                }
+            }
+        } else {
+            tmp
+        }
+    } else {
+        tmp
+    };
```
neonphog, line 75:
```suggestion
    } else {
        // We could check again here to see if the perf_path is correct.
        // Some other parallel process my have written it. Then delete
        // our temp file, which would be a slightly better custodian
        // of disk space.
        tmp
    };
```

Issue #39, opened 2023-05-28
Implement backend-webrtc-rs so we can compare performance with backend-go-pion


Connoropolous:
Trying to use `backend-webrtc-rs` instead of `backend-go-pion` but it won't compile. 

This is running `cargo build --release --target aarch64-apple-ios`, but maybe it also happens for normal targets. Not sure yet, haven't checked.

```bash
error[E0433]: failed to resolve: use of undeclared crate or module `imp`
  --> /Users/connor/.cargo/registry/src/github.com-1ecc6299db9ec823/tx5-0.0.1-alpha.17/src/back_buf.rs:13:10
   |
13 |     imp: imp::ImpWriter,
   |          ^^^ use of undeclared crate or module `imp`

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
  --> /Users/connor/.cargo/registry/src/github.com-1ecc6299db9ec823/tx5-0.0.1-alpha.17/src/back_buf.rs:22:18
   |
22 |             imp: imp::ImpWriter::new()?,
   |                  ^^^ use of undeclared crate or module `imp`

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
  --> /Users/connor/.cargo/registry/src/github.com-1ecc6299db9ec823/tx5-0.0.1-alpha.17/src/back_buf.rs:65:21
   |
65 |     pub(crate) imp: imp::Imp,
   |                     ^^^ use of undeclared crate or module `imp`

error[E0433]: failed to resolve: use of undeclared crate or module `tx5_go_pion`
  --> /Users/connor/.cargo/registry/src/github.com-1ecc6299db9ec823/tx5-0.0.1-alpha.17/src/back_buf.rs:76:33
   |
76 |     pub(crate) fn from_raw(buf: tx5_go_pion::GoBuf) -> Self {
   |                                 ^^^^^^^^^^^ use of undeclared crate or module `tx5_go_pion`

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
  --> /Users/connor/.cargo/registry/src/github.com-1ecc6299db9ec823/tx5-0.0.1-alpha.17/src/back_buf.rs:78:18
   |
78 |             imp: imp::Imp::from_raw(buf),
   |                  ^^^ use of undeclared crate or module `imp`

error[E0433]: failed to resolve: use of undeclared crate or module `imp`
[...]

PR #41, opened 2023-06-01
backend-webrtc-rs code

Connoropolous:
rough direction check for #39

### DIFF
--- a/crates/tx5/src/back_buf/imp/imp_webrtc_rs.rs
+++ b/crates/tx5/src/back_buf/imp/imp_webrtc_rs.rs
@@ -1,0 +1,115 @@
+use crate::*;
+use bytes::{Bytes, BytesMut, BufMut, Buf};
+
+pub struct ImpWriter {
+    buf: BytesMut,
+}
+
+impl ImpWriter {
+    #[inline]
+    pub fn new() -> Result<Self> {
+        Ok(Self {
+            buf: BytesMut::new(),
+        })
+    }
+
+    #[inline]
+    pub fn finish(self) -> Imp {
+        Imp { buf: self.buf }
+    }
+}
+
+impl std::io::Write for ImpWriter {
+    #[inline]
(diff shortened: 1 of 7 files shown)


crates/tx5/src/back_buf/imp/imp_webrtc_rs.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    #[inline]
+    fn flush(&mut self) -> std::io::Result<()> {
+        let mut writer = self.buf.writer();
+        writer.flush()
+    }
+}
+
+pub struct Imp {
+    pub(crate) buf: BytesMut,
```
neonphog, line 52: Would it work to have this one be a non-mut `Bytes`, and freeze in the `from_*` constructors? I don't *think* we ever mutate the backbufs and then the try_clones would be cheaper.

Connoropolous: yep, appears to work

Connoropolous: have I resolved?

Connoropolous:
@neonphog I've spent some more time on this...
I made some progress, but I can't figure out how to do the MultiEvt  / multiplexing thing that you did for the go implementation. There are some funky details that I see were required to manage the Rust/Go boundary in terms of callbacks, and I'm pretty sure it wouldn't make sense to re-implement those for a Rust only version

neonphog:
@Connoropolous 

> @neonphog I've spent some more time on this... I made some progress, but I can't figure out how to do the MultiEvt / multiplexing thing that you did for the go implementation. There are some funky details that I see were required to manage the Rust/Go boundary in terms of callbacks, and I'm pretty sure it wouldn't make sense to re-implement those for a Rust only version

Yep, totally, no worries. Thanks so much for your start on this, it'll give me a jump when I can get back to this :+1:

PR #45, opened 2023-08-01
Influxive

neonphog:


crates/tx5/src/state/test.rs
```diff
@@ -105,12 +124,19 @@ impl Test {
...
+            .query(
+                r#"from(bucket: "influxive")
+    |> range(start: -15m, stop: now())
+    "#,
+            )
+            .await
             .unwrap();
-        println!("{}", String::from_utf8_lossy(&buf));
+
+        println!("{result}");
```
ThetaSinner, line 137: Is there anything that can be asserted on `result` to check this automatically?

crates/tx5/src/endpoint.rs
```diff
@@ -383,6 +403,132 @@ async fn new_conn_task(
...
+    );
+    impl Drop for Unregister {
+        fn drop(&mut self) {
+            if let Some(mut unregister) = self.0.take() {
+                let _ = unregister.unregister();
+            }
+        }
+    }
+
+    let slot: Arc<std::sync::Mutex<Option<HashMap<String, BackendMetrics>>>> =
```
ThetaSinner, line 421: What is the lifetime of this? It looks like it's dropped after `on_new_conn` finishes and with the use of `Weak` neither the task nor the callback will keep it alive

ThetaSinner (approved):

crates/tx5/src/state/test.rs
neonphog: mmm... the state logic isn't really the right place for this. Since there's no real connection, no real metrics are actually collected. This was more about making sure there aren't any panics in the processing mechanisms.

I'll add a check for metrics in a higher level integration test.

neonphog: [done](https://github.com/holochain/tx5/pull/45/commits/105b9b631a545867722a28718d21205b39ee2c97) - unfortunately, due to the use of `opentelemetry_api::global`, I now have to run the tests single threaded... until I get around to implementing a test workaround for that.

crates/tx5/src/endpoint.rs
ThetaSinner: Oh I get it, there's a loop further down that I hadn't seen so this function doesn't exit until the connection ends. Thanks for the explanation

PR #48, opened 2023-08-14
ci: test rust 1.66.1 which is currently used by holochain; fixes for rust 1.66.1; add nix development shell

steveej:
please see the commit messages for details on the changes


Makefile
```diff
@@ -106,10 +106,7 @@ docs: tools
 tools: tool_rust tool_fmt tool_clippy tool_readme
 
 tool_rust:
-	@if rustup --version >/dev/null 2>&1; then \
-		echo "# Makefile # found rustup, setting override stable"; \
-		rustup override set stable; \
```
ThetaSinner, line 111: Oh this is interesting, I think I missed this in the Lair repo when I added 1.66.1 to CI

ThetaSinner (approved):

Makefile
neonphog: : / - i mean, i guess this is less necessary at the moment, since my system isn't defaulting to nightly, but it will be annoying not having it if I ever switch again.

flake.nix
```diff
@@ -0,0 +1,27 @@
+{
```
neonphog, line 1: I find this a bit invasive, like for a while years ago everyone was committing ide configs into repos until it became best practice ***not*** to do that.

neonphog (approved):

steveej:
thanks for the reviews!

@gregoryj note that influxive is currently broken in the nix shell so the tests that require it fail. i spent some time trying to get it to work with influxdb that's part of the nix shell, but influxive [currently hardcodes the version string it requires from the influx binaries (cli and daemon)](https://github.com/holochain/influxive/blob/102f19672975164e9b5a5c72ca7b6c77547e2516/crates/influxive-child-svc/src/lib.rs#L515-L519) and thus fails as the nixpkgs ones aren't the same.

neonphog:
@steveej 

We can relax the influxive versions a little if we need to. I just haven't yet had the time to determine how far back the current cli and protocol apis are valid.

Makefile
ThetaSinner: It really bothered me that this forces swapping out my toolchain while running unrelated commands. Especially as I'd picked a different toolchain on purpose to test the build against 1.66 which was replaced before the build actually happened. I'm fine with dockerfiles and nix files or whatever else is scoped to the project but I don't want my system state to be affected by running commands to build individual projects

neonphog: Mmm, I can see that. A convenience I'm not allowed to have, then.

PR #55, opened 2023-09-26
allow config of pion udp ephemeral port range

neonphog:

PR #58, opened 2023-09-26
Expose min/max port range for tx5 configuration

mattyg:
Thank you so much @neonphog -- this just adds onto your work to expose the `Tx5InitConfig` all the way up to the main tx5 `Config`, which I'll then add to the conductor's network tuning params config

### DIFF
--- a/crates/tx5/src/endpoint.rs
+++ b/crates/tx5/src/endpoint.rs
@@ -366,6 +366,7 @@ async fn new_conn_task(
     use tx5_go_pion::DataChannelEvent as DataEvt;
     use tx5_go_pion::PeerConnectionEvent as PeerEvt;
     use tx5_go_pion::PeerConnectionState as PeerState;
+    use tx5_go_pion::Tx5InitConfig;
 
     enum MultiEvt {
         OneSec,
@@ -383,12 +384,19 @@ async fn new_conn_task(
     let peer_snd2 = peer_snd.clone();
     let mut peer = match async {
         let peer_config = BackBuf::from_json(ice_servers)?;
+        let tx5_init_config = Tx5InitConfig {
+            ephemeral_udp_port_min: config.min_ephemeral_udp_port(),
+            ephemeral_udp_port_max: config.max_ephemeral_udp_port(),
+        };
 
-        let peer =
-            tx5_go_pion::PeerConnection::new(peer_config.imp.buf, move |evt| {
+        let peer = tx5_go_pion::PeerConnection::new(
+            peer_config.imp.buf,
+            tx5_init_config,
+            move |evt| {
                 let _ = peer_snd2.send(MultiEvt::Peer(evt));
-            })
-            .await?;
+            },
+        )
+        .await?;
 
         Result::Ok(peer)
     }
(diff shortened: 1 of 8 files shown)


mattyg:
the only other concern I have is setting the default min & max ports to 1 and 65355 - I believe the allowed ephemeral ports are different for every OS and wouldn't want pion to try to grab a port that is not permitted. I'll see if I can find what its is doing under the hood and if it's impacted.

crates/tx5/src/endpoint.rs
```diff
@@ -383,12 +384,19 @@ async fn new_conn_task(
...
+        let tx5_init_config = Tx5InitConfig {
+            ephemeral_udp_port_min: config.min_ephemeral_udp_port(),
+            ephemeral_udp_port_max: config.max_ephemeral_udp_port(),
+        };
 
-        let peer =
-            tx5_go_pion::PeerConnection::new(peer_config.imp.buf, move |evt| {
+        let peer = tx5_go_pion::PeerConnection::new(
+            peer_config.imp.buf,
+            tx5_init_config,
```
neonphog, line 394: It doesn't really make sense to me to have to pass the init config in every time if only the first one is going to matter. I was imagining the integration from the holochain side would be putting these ports in the kitsune tuning params, then in the kitsune actor which is only initialized once, it could directly:

```rust
tx5_go_pion::Tx5InitConfig { /* vals from tuning params */ }.set_as_global_default();
```

Before going on to create the tx5 Ep.

mattyg: ah okay great -- so this PR is unnecessary

mattyg: Though I'm not sure how we access tx5_go_pion in holochain if tx5 is supposed to abstract over a pion vs webrtc implementation?

neonphog: For reference, that could be right before this line: https://github.com/holochain/holochain/blob/91ee7b705ef256fbb05233dbdb928532fb301dc1/crates/kitsune_p2p/kitsune_p2p/src/spawn/actor/meta_net.rs#L845

It might make sense to re-export `Tx5InitConfig` from the `tx5` crate so we don't have to explicitly depend on `tx5_go_pion` separately.

neonphog: > Though I'm not sure how we access tx5_go_pion in holochain if tx5 is supposed to abstract over a pion vs webrtc implementation?

Ahh, good point... I wonder if `Tx5InitConfig` should be moved to `tx5_core` then?

neonphog: I'll see if that works out

mattyg: Do we need to expose some function in tx5 or tx5_core that calls the set_as_global_default() and has a feature flag for pion. While the same function under the webrtc-rs feature flag does nothing. Does that make sense?

mattyg: nevermind we can apply the same config to webrtc-rs https://github.com/webrtc-rs/webrtc/discussions/352

I'm guessing that's the right move

neonphog: Okay, hopefully this will make it easy to work with: https://github.com/holochain/tx5/pull/55/commits/30365912f364e0e8a78b4bd443499959a8b8b450

PR #60, opened 2023-10-08
Build/go disable buildvcs

mattyg:
This adds the flag `-buildvcs=false` to go build commands. I don't understand why, but running `cargo run` in a docker compose command fails due to an error thrown by this buildvcs feature.  Edit: I figured out a workaround to avoid this.

My vague understanding is that the buildvcs feature attaches your version control system's current commit hash onto the built binary. Not sure if this is useful, since we're just wrapping the binary within another anyway. If it is necessary let me know and I'll work around it. 

This is how the go manpage explains it:

```
-buildvcs
	Whether to stamp binaries with version control information
	("true", "false", or "auto"). By default ("auto"), version control
	information is stamped into a binary if the main package, the main module
	containing it, and the current directory are all in the same repository.
	Use -buildvcs=false to always omit version control information, or
	-buildvcs=true to error out if version control information is available but
	cannot be included due to a missing tool or ambiguous directory structure.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /target
 /output-*
+.cargo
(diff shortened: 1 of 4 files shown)


neonphog (approved):
I agree it seems correct to set this to false.

PR #68, opened 2023-10-26
remove workaround to test rustix fix for tempfile::persist_noclobber

neonphog:
https://github.com/Stebalien/tempfile/issues/259

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,13 +146,14 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.3"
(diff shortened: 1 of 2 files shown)


neonphog:
@guillemcordoba - FYI - As the bug that was causing that flags empty assertion on android has now been fixed upstream, I've removed the workaround here. As this makes its way into code you're using, you'll need to ensure your Cargo.lock files are updated to use rustix >= `0.38.21` where the fix was made. Thanks!

guillemcordoba:
Thank you!

PR #72, opened 2023-11-28
Async conn events

neonphog:
- [x] fix failing tests deadlock

Tests passing with a workaround... going to not merge this by itself, but instead along with the state update because there is a chance this is strictly less correct as-is. With the design of the coming state work, the incorrectness should be impossible.


neonphog:
The work here was originally included in https://github.com/holochain/tx5/pull/77 - but largely had to be removed because of the go pion deadlock when closing things during event callbacks.

PR #73, opened 2023-11-30
Fix cross-compiling from darwin to ios

guillemcordoba:
Fixes https://github.com/holochain/tx5/issues/71

Based on https://github.com/jedisct1/libsodium-sys-stable/blob/master/build.rs

### DIFF
--- a/crates/tx5-go-pion-sys/build.rs
+++ b/crates/tx5-go-pion-sys/build.rs
@@ -155,8 +155,7 @@ fn go_build_cmd(
 
     cmd.env("CGO_ENABLED", "1");
 
-    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
-    if target_os == "ios" {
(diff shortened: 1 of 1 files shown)


crates/tx5-go-pion-sys/build.rs
```diff
@@ -155,6 +155,29 @@ fn go_build_cmd(
 
     cmd.env("CGO_ENABLED", "1");
 
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
+    if target_os == "ios" {
```
neonphog, line 159: Can you use the already determined TARGET.go_os?

```suggestion
    if TARGET.go_os == "ios" {
```

neonphog (changes requested):
Awesome! Thanks for finding this :rocket:

neonphog (approved):

PR #76, opened 2023-12-15
build: bump lair_keystore_api

jost-s:
Bumping lair_keystore_api in this package, because it doesn't allow `0.4.0` with the current version requirement and pulls in `rmp-serde` indirectly. Do you prefer `>=0.3.0` or `0.4.0`?
`>=0.3.0` allows for anything greater or equal to 0.3.0, major, minor and patch
`0.4.0` is `>=0.4.0 and <0.5.0`


neonphog:
@jost-s - Not sure I have a strong opinion. Technically this time it's probably okay since the actual api didn't change, but if version 0.5 no longer compiles as-is, it's probably safer to use "0.4.0" right?

jost-s:
Yes, I agree.

Issue #69, opened 2023-10-31
Can't write to `/data/local/tmp` on Android API >= 30


guillemcordoba:
This is more my way of explaining everything I've learned in my latest explorations trying to run holochain in android, than an issue in this library, although _there is_ an error that I get in my Redmi Note 7 when running it.

So, it seems that from Android 11 (API level 30) on, Android tightened security around file system reads and writes, and [doesn't allow applications to just write to `/data/local/tmp`](https://stackoverflow.com/questions/23424602/android-permission-denied-for-data-local-tmp). This is the error I get when running tx5 on android, I get it [from this line](https://github.com/holochain/tx5/blob/40285ab7b3ea547df8414396f84f095b4a9b70e5/crates/tx5-core/src/file_check.rs#L132). 

I was trying to override the [result for this line](https://github.com/holochain/tx5/blob/40285ab7b3ea547df8414396f84f095b4a9b70e5/crates/tx5-core/src/file_check.rs#L29C6-L29C6) by forking `dirs-rs` and redirecting the `get_local_dir` call to [`app_dirs2` get_app_dir](https://github.com/app-dirs-rs/app_dirs2/blob/ffce45b500bdb1a7324e8c3267371e2ffe452221/src/imp/platform/android.rs#L60), but I had no luck with this, it complained that the file didn't exist [in this line](https://github.com/holochain/tx5/blob/40285ab7b3ea547df8414396f84f095b4a9b70e5/crates/tx5-core/src/file_check.rs#L97) (are we missing a `.create(true)` in that line??). I tried to add the `.create(true)`, but then I bumped into another error: `InvalidArgument`, from that same function, but not sure which line...

After this, I discovered [this issue](https://github.com/Stebalien/tempfile/issues/235#issuecomment-1520479951), and that worked! Seems that by setting the `TMPDIR` env var, we can change where the `tempdir` crate creates new files. So, I can now compile and run holochain in android, by adding this workaround to my initialization code: 

```rust
    let dir = app_dirs2::app_root(
        app_dirs2::AppDataType::UserCache,
        &app_dirs2::AppInfo {
            name: "studio.darksoil.rostangatest",
            author: "darksoil.studio",
        },
    )?;
    std::env::set_var("TMPDIR", dir);
```

After this, no more `permission denied` error, and networking works! I am able to communicate with a holochain node running in my laptop, and fetches for records work correctly! So this is really exciting.

[...]

PR #77, opened 2023-12-19
Refactor of endpoint and connection management state logic

neonphog:
- The main change of this PR is an inversion of the async driver for a tx5 endpoint.
  - Previously a central driver task would attempt to control all aspects, but the edge cases proved difficult to both reason about and test.
  - Now the main async driver is actually the `send` function. If we don't have a signal connection it will trigger creation of a cloneable future and await, following if there is no peer connection it will trigger creation of a cloneable peer connection future and await. The bulk of the complexity moves to ensuring these futures are cleaned up appropriately in error conditions.


crates/tx5-core/src/evt.rs
```diff
@@ -0,0 +1,105 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EventPermit").finish()
+    }
+}
+
+/// Generic event sender.
+pub struct EventSend<E: From<Error>> {
+    limit: Arc<tokio::sync::Semaphore>,
+    send: tokio::sync::mpsc::UnboundedSender<(E, EventPermit)>,
+}
```
neonphog, line 17: Explicitly bound channel, lets us send bounded (backpressured) data normally, but unbounded errors.

crates/tx5-core/src/file_check.rs
```diff
@@ -17,6 +17,19 @@ impl FileCheck {
...
+        app_dirs2::AppDataType::UserCache,
+        &app_dirs2::AppInfo {
+            name: "host.holo.tx5",
+            author: "host.holo.tx5",
+        },
+    ) {
+        Ok(dir) => Some(dir),
+        _ => None,
+    }
+}
```
neonphog, line 31: Changes in this file reflect fixes for https://github.com/holochain/tx5/issues/69

crates/tx5-go-pion/src/evt.rs
```diff
@@ -56,150 +58,282 @@ pub enum PeerConnectionEvent {
...
-            }
-            SysEvent::DataChanBufferedAmountLow(data_chan_id) => {
-                let maybe_cb =
-                    MANAGER.lock().data_chan.get(&data_chan_id).cloned();
-                if let Some(cb) = maybe_cb {
-                    cb(DataChannelEvent::BufferedAmountLow);
-                }
-            }
+        API.on_event(move |sys_evt| {
+            let _ = send.blocking_send(sys_evt);
```
neonphog, line 186: We're using this channel to translate from the synchronous golang FFI events into asynchronous rust. While this represents a bottleneck, the `manager_access!` macro should start killing off connections if the app is too slow to process incoming events within 18ms.

crates/tx5/src/ep3.rs
```diff
@@ -0,0 +1,600 @@
...
+    }
+}
+
+/// A signal server url.
+pub type SigUrl = Tx5Url;
+
+/// A peer connection url.
+pub type PeerUrl = Tx5Url;
+
+type SigMap = HashMap<SigUrl, (u64, AbortableTimedSharedFuture<Arc<Sig>>)>;
```
neonphog, line 87: This `AbortableTimedSharedFuture` is the core enabling mechanic of this refactor, you can find the definition in `crates/tx5/src/lib.rs`. The first thread/task to ask for a particular signal server will add an entry to this hash map, then all future requests will await on that same future. The complexity comes in cleaning these up in the case of errors.

crates/tx5/src/ep3.rs
```diff
@@ -0,0 +1,600 @@
...
+        }
+        Ok(r) => Ok(r),
+    }
+}
+
+fn close_sig(
+    weak_sig_map: &Weak<Mutex<SigMap>>,
+    sig_url: &SigUrl,
+    close_sig_uniq: u64,
+) {
```
neonphog, line 568: This is the entrypoint for cleaning up the shared futures in the hash map referenced in my previous comment.

crates/tx5/src/lib.rs
```diff
@@ -36,261 +36,177 @@ pub mod deps {
...
+/// Make a shared (clonable) future abortable and set up an automatic
+/// abort at a time in the future specified by a duration.
+struct AbortableTimedSharedFuture<T: Clone> {
+    f: futures::future::Shared<
+        futures::future::BoxFuture<'static, std::result::Result<T, Error>>,
+    >,
+    a: std::sync::Arc<
+        std::sync::Mutex<Option<tokio::sync::oneshot::Sender<Error>>>,
+    >,
+    t: DoDrop,
```
neonphog, line 77: Here's the definition of the `AbortableTimedSharedFuture`.

crates/tx5/src/proto.rs
```diff
@@ -0,0 +1,580 @@
+//! Types associated with the Tx5 protocol.
```
neonphog, line 1: The tx5 protocol solves 2 problems at once:

- First, webrtc only supports messages up to 16K, so this lets us send bigger messages.
- Second, *if* we're sending bigger messages, we have to worry about the size in memory taken up by the receiving side. This protocol lets us request permits to send larger messages, which gives the receiving side a tool to be able to manage more open connections at the same time without worrying about the worst case memory usage of each connection all at the same time.

crates/tx5/src/ep3/test.rs
```diff
@@ -0,0 +1,623 @@
...
+    }
+
+    println!("{sort:?}");
+
+    // make sure the there is no cross-messaging
+    for (_, list) in sort.iter() {
+        let (check_node_id, _, _) = list.get(0).unwrap();
+
+        for (node_id, _, _) in list.iter() {
+            //println!("{check_node_id}=={node_id}");
```
jost-s, line 390:
```suggestion
```

jost-s (approved):
LGTM, with many tests and simple code with a few exceptions

crates/tx5-go-pion/src/data_chan.rs
```diff
@@ -1,106 +1,209 @@
...
+        f.write_str(&res)
     }
 }
 
 impl DataChannel {
+    pub(crate) fn new(
+        data_chan_id: usize,
+        recv_limit: Arc<tokio::sync::Semaphore>,
+    ) -> (Self, EventRecv<DataChannelEvent>) {
+        let (evt_send, evt_recv) = EventSend::new(1024);
```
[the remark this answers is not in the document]
neonphog: I don't think it makes sense that individuals would need a different value here. We may identify that this value is sub-optimal, but if we make a change it would benefit everybody, so if they had it configured to something specific they would not receive that benefit.

crates/tx5/src/ep3/test.rs
```diff
@@ -0,0 +1,623 @@
...
+    let (_cli_url1, ep1, _ep1_recv) = test.ep(config.clone()).await;
+    let (cli_url2, _ep2, mut ep2_recv) = test.ep(config).await;
+
+    ep1.send(cli_url2, b"hello").await.unwrap();
+
+    let res = ep2_recv.recv().await.unwrap();
+    match res {
+        Ep3Event::Connected { .. } => (),
+        _ => panic!(),
+    }
```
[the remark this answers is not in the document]
neonphog: For the most part I find actual match statements more readable, and easier to update if it comes up that you need to handle multiple arms differently.

crates/tx5/src/proto.rs
maackle: This would be good as a code or README comment

maackle (commented):
More high-level comments would be pretty helpful for this PR and for future generations. The things you felt called to make PR comments on could be code comments too. Also, mod-level comments would be helpful, especially for the new files in this PR.

crates/tx5/src/lib.rs
```diff
@@ -36,261 +36,177 @@ pub mod deps {
...
-        *self & !FINISH
-    }
+#[derive(Clone)]
+struct DoDrop(std::sync::Arc<tokio::task::JoinHandle<()>>);
 
-    fn is_finish(&self) -> bool {
-        *self & FINISH > 0
+impl Drop for DoDrop {
+    fn drop(&mut self) {
+        self.0.abort();
```
maackle, line 54: Has tokio task cancellation always been a thing or is it a recent thing? I always remember tasks being uncancelable.

crates/tx5/src/lib.rs
```diff
@@ -36,261 +36,177 @@ pub mod deps {
...
+        Self {
+            f: futures::future::FutureExt::shared(
+                futures::future::FutureExt::boxed(async move {
+                    tokio::select! {
+                        r = async {
+                            Err(ar.await.map_err(|_| Error::id("AbortHandleDropped"))?)
+                        } => r,
+                        r = f => r,
+                    }
+                }),
```
maackle, line 119: I wonder if plain ol' tokio::time::timeout would have worked too? Just a task that awaits the timeout of the input future and maps the timeout error to the custom error? This is fine too but seems like a reinvention of timeout?

crates/tx5/src/lib.rs
```diff
@@ -36,261 +36,177 @@ pub mod deps {
...
+            let a = a.clone();
+            tokio::task::spawn(async move {
+                tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+                a.abort(Error::id("abort").into());
+            });
         }
+        let r = a.await;
+        assert_eq!("abort", r.unwrap_err().to_string());
     }
 }
```
maackle, line 212: Is there a way to test the drop behavior? Is it important?

crates/tx5/src/proto.rs
```diff
@@ -0,0 +1,580 @@
...
+///
+/// The initial 3 bits representing values 0, 1, and 7 are reserved.
+/// Decoders should error on receiving these values.
+#[derive(Debug, PartialEq)]
+pub(crate) enum ProtoHeader {
+    /// This is a protocol version handshake.
+    /// `3 bits == 2`
+    /// `5 bits == version number (currently 1)`
+    /// `24 bytes == as bytes b"tx5"`
+    Version(u8, [u8; 3]),
```
maackle, line 23: I'm confused about how to interpret this. 3 * 8 = 24, but this seems to add up to 3 + 5 + 24 = 32?

crates/tx5/src/proto.rs
```diff
@@ -0,0 +1,580 @@
...
+/// 4 byte Tx5 protocol header.
+///
+/// The initial 3 bits representing values 0, 1, and 7 are reserved.
+/// Decoders should error on receiving these values.
+#[derive(Debug, PartialEq)]
+pub(crate) enum ProtoHeader {
+    /// This is a protocol version handshake.
+    /// `3 bits == 2`
+    /// `5 bits == version number (currently 1)`
+    /// `24 bytes == as bytes b"tx5"`
```
maackle, line 22:
```suggestion
    /// `24 bits == as bytes b"tx5"`
```

crates/tx5/src/proto.rs
```diff
@@ -0,0 +1,580 @@
...
+            3 => Ok(Self::CompleteMessage(r.get_bits(..29))),
+            4 => Ok(Self::MultipartMessage(r.get_bits(..29))),
+            5 => Ok(Self::PermitRequest(r.get_bits(..29))),
+            6 => Ok(Self::PermitGrant(r.get_bits(..29))),
+            _ => Err(Error::id("ReservedHeaderBits")),
+        }
+    }
+
+    /// Encode a Tx5 protocol header into canonical 4 bytes.
+    pub fn encode(&self) -> Result<(u8, u8, u8, u8)> {
```
maackle, line 70: This seems easier to work with but maybe it's just a matter of preference

```suggestion
    pub fn encode(&self) -> Result<[u8; 4]> {
```

crates/tx5/src/proto.rs
```diff
@@ -0,0 +1,580 @@
...
+
+/// Tx5 protocol decoder.
+pub(crate) struct ProtoDecoder {
+    state: DecodeState,
+    want_size: usize,
+    incoming: Vec<u8>,
+    want_remote_grant: bool,
+    did_error: bool,
+    grant_permit: Option<tokio::sync::OwnedSemaphorePermit>,
+    grant_notify: Option<tokio::sync::oneshot::Sender<()>>,
```
maackle, line 218: I highly recommend folding as much of this extra state as possible into the `DecodeState` enum. It looks like most or all of it is correlated: `want_size`, `incoming`, and `grant_permit` all seem only relevant to `RemoteAwaitingPermit` and could be fields on that variant. It's so helpful when working with state machines to keep them constrained as possible, for so many reasons.

The ideal would be that `ProtoDecoder` and `DecodeState` are the same thing.

crates/tx5/src/proto.rs
```diff
@@ -0,0 +1,580 @@
...
+            check(ProtoHeader::CompleteMessage(*v));
+            check(ProtoHeader::MultipartMessage(*v));
+            check(ProtoHeader::PermitRequest(*v));
+            check(ProtoHeader::PermitGrant(*v));
+        }
+    }
+
+    #[test]
+    fn proto_header_overflow() {
+        assert!(ProtoHeader::Version(u8::MAX, [b't', b'x', b'5'])
```
maackle, line 407:
```suggestion
        assert!(ProtoHeader::Version(0b00100000000000000000000000000000, [b't', b'x', b'5'])
```

crates/tx5/src/lib.rs
neonphog: [It's been around at least since tokio 1.0](https://docs.rs/tokio/1.0.1/tokio/task/struct.JoinHandle.html#method.abort)

crates/tx5-core/src/evt.rs
```diff
@@ -0,0 +1,105 @@
...
+/// Permit for sending on the channel.
+pub struct EventPermit(Option<tokio::sync::OwnedSemaphorePermit>);
+
+impl std::fmt::Debug for EventPermit {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EventPermit").finish()
+    }
+}
+
+/// Generic event sender.
```
maackle, line 13:
```suggestion
/// Generic event sender.
///
/// The sender is backed by an unbounded channel, but a bound is enforced
/// by a semaphore to allow bounds on normal data but unbounded errors.
```

crates/tx5-go-pion/src/data_chan.rs
```diff
@@ -1,106 +1,209 @@
...
+
+        let strong = Arc::new(Mutex::new(Ok(DataChanCore::new(
```
maackle, line 114: How about this at least?

```suggestion

    const CAPACITY: usize = 1024;

    pub(crate) fn new(
        data_chan_id: usize,
        recv_limit: Arc<tokio::sync::Semaphore>,
    ) -> (Self, EventRecv<DataChannelEvent>) {
        let (evt_send, evt_recv) = EventSend::new(Self::CAPACITY);

        let strong = Arc::new(Mutex::new(Ok(DataChanCore::new(
```

crates/tx5-go-pion/src/evt.rs
```diff
@@ -56,150 +59,444 @@ pub enum PeerConnectionEvent {
...
+        // so that the go thread trying to send the event is blocked
+        let _ = r.blocking_recv();
+    }
+}
 
-pub(crate) type DataChanEvtCb =
-    Arc<dyn Fn(DataChannelEvent) + 'static + Send + Sync>;
+macro_rules! manager_access {
+    ($man:ident, $id:ident, $off:ident, $map:ident, $code:expr) => {
+        let $map = $man.lock().unwrap().$map.get(&$id).cloned();
```
maackle, line 246: See man map. Map, man, map!

crates/tx5-go-pion/src/evt.rs
```diff
@@ -56,150 +59,444 @@ pub enum PeerConnectionEvent {
...
-                }
+                let mut buf = GoBuf(buffer_id);
+                manager_access!(
+                    MANAGER,
+                    data_chan_id,
+                    offload,
+                    data_chan,
+                    async {
+                        let len = buf.len()?;
+                        if len > 16 * 1024 {
```
maackle, line 363: Is this the same as the EventSend capacity, or something different? Names and/or intentions would be great for these constants.

maackle (commented):
Monumental! Looks pretty good, nice tests too. My comments are mainly about comments.

crates/tx5/src/ep3.rs
```diff
@@ -0,0 +1,609 @@
...
+}
+
+impl Default for Config3 {
+    fn default() -> Self {
+        Self {
+            connection_count_max: 4096,
+            send_buffer_bytes_max: 64 * 1024,
+            recv_buffer_bytes_max: 64 * 1024,
+            incoming_message_bytes_max: 512 * 1024 * 1024,
+            message_size_max: 16 * 1024 * 1024,
```
ThetaSinner, line 145: What's the difference between the message side and the incoming message size? Just looking at this it seems unnecessary to allow more incoming than should be sent?

crates/tx5/src/ep3.rs
```diff
@@ -0,0 +1,609 @@
...
+        // so weak persistence security is okay,
+        // since it will not be persisted.
+        // The private keys will still be mem_locked
+        // so they shouldn't be swapped to disk.
+        let keystore_config = PwHashLimits::Minimum
+            .with_exec(|| LairServerConfigInner::new("/", passphrase.clone()))
+            .await
+            .unwrap();
+
+        let _lair_keystore = PwHashLimits::Minimum
```
ThetaSinner, line 254: This gets used, does it need to be `_` prefixed?

ThetaSinner (commented):
This looks good to me, I'd need a lot more time to dig into the Go code and protocol to understand all the details and feel like I'd done a really good review. I've asked some questions to try and learn a bit, hope that's okay!

crates/tx5-core/src/evt.rs
neonphog: done

crates/tx5/src/proto.rs
neonphog: done

crates/tx5-go-pion/src/data_chan.rs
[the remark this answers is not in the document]
neonphog: k, made it configurable

crates/tx5-core/src/error.rs
```diff
@@ -8,6 +8,11 @@ pub struct Error {
     pub info: String,
 }
 
+impl From<Error> for () {
+    #[inline]
+    fn from(_: Error) -> Self {}
+}
+
```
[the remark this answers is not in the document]
neonphog: Yes, I sometimes use `Result<_, ()>` in unit tests. I don't think there's any way to accidentally loose information unless we have the `Result<_, ()>` type in actual code, which we don't. I'm inclined to leave this in here for my convenience, but could be convinced otherwise if you feel strongly about it.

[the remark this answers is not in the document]
maackle: I couldn't find any instances of that in tx5. If it's for tests only and limited to the core crate, please put a `#[cfg(test)]` on it.

I'd like to see examples of how you use it. I can't fathom why it would be helpful to throw away an error, even during a test, since if a test fails you want to see why.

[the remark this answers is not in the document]
neonphog: *sigh* no convenience... I can delete it.

crates/tx5/src/proto.rs
neonphog: I understand the desire to do this, but I've found the reality in rust doesn't actually work. When you match on the state it captures a reference to self, and makes it almost impossible to modify that state (switch to a different enum variant) while you're matching on it.

The complexity bloomed, and I had to pull in things like [replace_with](https://crates.io/crates/replace_with).

crates/tx5-core/src/error.rs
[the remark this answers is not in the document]
neonphog: done: https://github.com/holochain/tx5/pull/77/commits/95210064c7afde25dad5afbde55ae1c8a65d2f00

crates/tx5/src/proto.rs
neonphog: done, but it was a u8: https://github.com/holochain/tx5/pull/77/commits/7608bfa10c4e5abef44ff55bee324c9e389ec590

crates/tx5-core/src/evt.rs
neonphog: Fixed this per your other comment

crates/tx5-go-pion/src/data_chan.rs
neonphog: This has been made into a config field.

crates/tx5-go-pion/src/evt.rs
neonphog: This is a core limit of the webrtc protocol itself, so there's no need to test it... but I can turn it into a constant at least.

neonphog: done: https://github.com/holochain/tx5/pull/77/commits/430e250b4604163d5ea183b15e5ba10919cd6f49

crates/tx5-core/src/evt.rs
```diff
@@ -0,0 +1,105 @@
...
+        self.0.recv().await.map(|r| r.0)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn event_limit() {
```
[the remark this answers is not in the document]
neonphog: Yep, you got it.

crates/tx5-core/src/evt.rs
```diff
@@ -0,0 +1,105 @@
...
+
+impl<E: From<Error>> EventRecv<E> {
+    /// Receive incoming PeerConnection events.
+    pub async fn recv(&mut self) -> Option<E> {
+        self.0.recv().await.map(|r| r.0)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
neonphog: I'm not sure those tests would provide any value. They aren't async functions, so you can't await them... And if you pull the items off the recv side you implicitly make the permits free. I assert that the send_permit and send_err functions are correct if they compile.

crates/tx5-core/src/file_check.rs
```diff
@@ -17,6 +17,19 @@ impl FileCheck {
     }
 }
 
+fn get_loc_dir() -> Option<std::path::PathBuf> {
```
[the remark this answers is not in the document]
neonphog: done: https://github.com/holochain/tx5/pull/77/commits/599a0511458c33b903f9edc1b8b8518d0c4e8e4b

crates/tx5-go-pion/src/data_chan.rs
```diff
@@ -1,106 +1,209 @@
...
-pub struct DataChannelSeed(pub(crate) usize, Arc<Mutex<Vec<DataChannelEvent>>>);
+pub(crate) struct DataChanCore {
+    data_chan_id: usize,
+    recv_limit: Arc<tokio::sync::Semaphore>,
+    evt_send: EventSend<DataChannelEvent>,
+    drop_err: Error,
+}
 
-impl Drop for DataChannelSeed {
+impl Drop for DataChanCore {
```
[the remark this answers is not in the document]
neonphog: It's actually more about *us* knowing a data channel is closed, and attempting to only emit that event once. Previously, if a channel was explicitly closed with an error, we'd get that error, and then also a following channel dropped error.

crates/tx5-go-pion/src/data_chan.rs
```diff
@@ -1,106 +1,209 @@
...
+    pub fn close(&self, err: Error) {
+        if let Some(strong) = self.0.upgrade() {
+            let mut tmp = Err(err.clone());
+
+            {
+                let mut lock = strong.lock().unwrap();
+                if let Ok(core) = &mut *lock {
+                    core.close(err);
+                }
+                std::mem::swap(&mut *lock, &mut tmp);
```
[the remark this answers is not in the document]
neonphog: done: https://github.com/holochain/tx5/pull/77/commits/84c519caf3d2e8057c73d3712f33f18d538b29c4

crates/tx5-go-pion/src/data_chan.rs
```diff
@@ -1,106 +1,209 @@
...
+
+        register_data_chan(data_chan_id, weak);
+
+        // we might have missed some state callbacks
+        if let Ok(ready_state) =
+            unsafe { API.data_chan_ready_state(data_chan_id) }
+        {
+            // this is a terrible suggestion clippy
+            #[allow(clippy::comparison_chain)]
+            if ready_state == 2 {
```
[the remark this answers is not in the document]
neonphog: Indeed, they are defined in go: https://pkg.go.dev/github.com/pion/webrtc/v3#DataChannelState

I don't disagree that it'd be nicer to have a rust enum... I did that once: https://docs.rs/tx5-go-pion/0.0.6-alpha/tx5_go_pion/enum.PeerConnectionState.html

But it's literally truth duplication...

crates/tx5-go-pion/src/evt.rs
```diff
@@ -56,150 +59,444 @@ pub enum PeerConnectionEvent {
...
     Close,
 
-    /// DataChannel incoming message.
-    Message(GoBuf),
+    /// Received incoming message on the data channel.
+    Message(GoBuf, tokio::sync::OwnedSemaphorePermit),
 
-    /// DataChannel buffered amount is now low.
+    /// Data channel buffered amount is now low.
     BufferedAmountLow,
```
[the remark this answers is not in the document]
neonphog: Actually, it's the opposite. It means our outgoing buffer was previously over the notification threshold an now has dropped below it, signaling it's ready to receive new outgoing data.

crates/tx5-go-pion/src/evt.rs
```diff
@@ -56,150 +59,444 @@ pub enum PeerConnectionEvent {
...
+            data_chan: HashMap::new(),
+        })
+    }
+
+    pub fn register_peer_con(
+        &mut self,
+        id: usize,
+        peer_con: peer_con::WeakPeerCon,
+    ) {
+        self.peer_con.insert(id, peer_con);
```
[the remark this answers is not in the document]
neonphog: There is one "endpoint" for all of kitsune, and the id is associated with the endpoint. Unless someone hacks their node, there should only be one connection between conductors across all spaces.

But if you run multiple conductors on a single machine, they should have different IDs, and thus multiple connections.

crates/tx5/src/ep3.rs
```diff
@@ -0,0 +1,609 @@
...
+use futures::future::BoxFuture;
+use lair_keystore_api::prelude::*;
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex, Weak};
+use tx5_core::{Error, EventRecv, EventSend, Id, Result, Tx5Url};
+
+fn next_uniq() -> u64 {
+    static UNIQ: std::sync::atomic::AtomicU64 =
+        std::sync::atomic::AtomicU64::new(1);
+    UNIQ.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
```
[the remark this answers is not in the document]
neonphog: These results are required to be unique, it doesn't matter if they come out-of-order, so Relaxed is correct.

crates/tx5/src/ep3.rs
neonphog:
```rust
    /// Maximum receive message reconstruction bytes in memory
    /// (accross entire endpoint). Default 512 MiB.
    pub incoming_message_bytes_max: u32,

    /// Maximum size of an individual message. Default 16 MiB.
    pub message_size_max: u32,
```

The `incoming_message_bytes_max` is our total buffer for reconstructing multi-part/chunked messages. It is shared across all connections, and allows us to keep many connections open without worrying about their memory usage.

The `message_size_max` is literally the biggest message we allow to be sent to us.

crates/tx5/src/ep3.rs
neonphog: It's not used once it goes into the struct, and I like:

```rust
Self {
    _lair_keystore,
```

better than:

```rust
Self {
    _lair_keystore: lair_keystore,
```

crates/tx5-go-pion/src/data_chan.rs
```diff
@@ -1,106 +1,209 @@
...
+
+        (Self(strong), evt_recv)
+    }
+
+    /// Close this data channel.
+    pub fn close(&self, err: Error) {
+        let mut tmp = Err(err.clone());
+
+        {
+            let mut lock = self.0.lock().unwrap();
```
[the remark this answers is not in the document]
neonphog: added the swap check here: https://github.com/holochain/tx5/pull/77/commits/6cc8ace3bcaa537e181bb8ebad1ac94a024a5224

I don't disagree about the unit tests, except it's super hard to write tests for this low-level crate, because there are so many fiddly bits, you end up having to duplicate a ton of the connection management code in the higher-level tx5 crate. The main current test is in the lib.rs of this crate.

crates/tx5/src/lib.rs
neonphog: done: https://github.com/holochain/tx5/pull/77/commits/4a24d8b6c0eff1218834dd8d0dd4971ed051602f

Yeah, I added the timeout bit after just having sorted out the aborting logic, so since I was holding a hammer, I hammered it... tokio::time::timeout is much better, thanks for pointing that out.

Also, I'm pretty sure the drop logic was just plain wrong. If one holochain task wanted to send some data but then got aborted, the drop logic would have aborted any future tasks that had tried to send data to the same endpoint even though they might have still wanted to send their data... so I got rid of that.

crates/tx5/src/lib.rs
```diff
@@ -36,261 +36,177 @@ pub mod deps {
...
 
-/*
-impl From<std::collections::VecDeque<bytes::Bytes>> for BytesList {
-    #[inline(always)]
-    fn from(v: std::collections::VecDeque<bytes::Bytes>) -> Self {
-        Self(v)
-    }
+/// Make a shared (clonable) future abortable and set up an automatic
+/// abort at a time in the future specified by a duration.
```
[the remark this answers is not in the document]
neonphog: Added additional comments here: https://github.com/holochain/tx5/pull/77/commits/4a24d8b6c0eff1218834dd8d0dd4971ed051602f

crates/tx5/src/lib.rs
neonphog: Drop logic was removed here: https://github.com/holochain/tx5/pull/77/commits/4a24d8b6c0eff1218834dd8d0dd4971ed051602f

crates/tx5-signal-srv/src/server.rs
```diff
@@ -277,7 +300,11 @@ async fn client_task(
                     }) => {
                         // TODO - pay attention to demo config flag
                         //        right now we just always honor demos
-                        srv_hnd_read.broadcast(msg);
+                        if let Some(srv_cmd) = srv_cmd_read.upgrade() {
+                            srv_cmd.broadcast(msg);
+                        } else {
+                            break;
```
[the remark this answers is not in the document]
neonphog: I wouldn't say so, no... In this else case we were not able to upgrade the weak reference, so that means whoever owns us has dropped the strong handle. If they want to error when that handle is dropped, they should make a newtype with a Drop impl, no?

crates/tx5-signal-srv/src/server.rs
```diff
@@ -286,19 +313,23 @@ async fn client_task(
...
+                                }
+                                Err(err) => {
+                                    tracing::trace!(?err);
                                 }
                             }
-                            Err(err) => {
-                                tracing::trace!(?err);
-                            }
+                        } else {
+                            break;
```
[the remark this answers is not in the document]
neonphog: see prev comment

crates/tx5/src/proto.rs
neonphog: For symmetry with the decode side, I'd prefer to keep them separate... You'd *think* you'd want `decode(data: [u8; 4])` BUT, that means you have to:

```rust
fn my_fn(in: Vec<u8>) {
    assert!(in.len() >= 4);
    decode([in[0], in[1], in[2], in[3]])
```

because rust doesn't have any straight forward way to use a Vec or `&[u8]` as a `[u8; 4]`...

crates/tx5/src/lib.rs
maackle: Looks good!

crates/tx5/src/proto.rs
maackle: Well there is try_into which is pretty straightforward: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3fb3a2f346241d2012a6d1a033385017

crates/tx5/src/proto.rs
maackle: Yeah, I have encountered that too...quite the bummer!

maackle (approved):
woohoo!

crates/tx5-go-pion/src/data_chan.rs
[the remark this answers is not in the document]
ThetaSinner: Fair enough, the main thing is that they're defined somewhere readily available. Good enough for me

crates/tx5-go-pion/src/evt.rs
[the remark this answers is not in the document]
ThetaSinner: Thanks! Makes sense

crates/tx5-signal-srv/src/server.rs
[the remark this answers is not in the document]
ThetaSinner: I see, yes that seems sensible!

ThetaSinner (approved):
Thanks for answering all my questions! Hopefully I contributed something as well as just learning :D PR looking good to me

PR #81, opened 2024-02-22
`musl-gcc` support

ThetaSinner:
This is to support building portable binaries when tx5 is a dependency. I've got everything else I need working except for strange linker errors against tx5 because it was building with the default Go C toolchain rather than the C toolchain I was using for the rest of the code. Depending on this branch results in a successful compile.

### DIFF
--- a/crates/tx5-go-pion-sys/build.rs
+++ b/crates/tx5-go-pion-sys/build.rs
@@ -339,7 +339,10 @@ fn gen_rust_const() {
 }
 
 fn find_musl_gcc() -> std::path::PathBuf {
-    let output = Command::new("which").arg("musl-gcc").output().expect("musl-gcc not found");
+    let output = Command::new("which")
+        .arg("musl-gcc")
+        .output()
+        .expect("musl-gcc not found");
 
     let s = String::from_utf8_lossy(&output.stdout);
     let s = s.trim();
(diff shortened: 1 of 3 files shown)


ThetaSinner:
Note that this won't work in the Nix environment yet, I'm working just with my own Cargo and testing the build of this in Holochain. I'll come back around to check that works okay when I'm done in Holochain.

crates/tx5-go-pion-sys/build.rs
```diff
@@ -319,3 +337,14 @@ fn gen_rust_const() {
         std::fs::write(&out, out_lines.join("\n")).unwrap();
     }
 }
+
+fn find_musl_gcc() -> std::path::PathBuf {
+    let output = Command::new("which")
+        .arg("musl-gcc")
+        .output()
+        .expect("musl-gcc not found");
```
neonphog, line 345: pretty linux specific, but i guess musl is by and large only used on linux anyways...

neonphog (approved):

crates/tx5-go-pion-sys/build.rs
ThetaSinner: Yes I'm currently only focused on `glibc` independence so binaries can be moved around between Linux distros that aren't on the very latest glibc like Nix is. I don't think macos has the same problem, we already fetch generic versions of libraries to work against for that... but I'll have to see when I get to that.

PR #82, opened 2024-03-18
various tx5 fixes

neonphog:
- support multiple iceservers blocks in `turn_doctor`
- account for possibility of multiple `Open` events on data channel in `turn-stress`
- fix multiple `sodoken` versions creeping in, and add a ci check to prevent that

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -87,9 +87,9 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.8.3"
+version = "0.8.4"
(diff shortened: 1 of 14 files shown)


steveej (approved):
i ran this successfully

```
  ( set -e;
    for i in `seq 1 10`; do
      cargo run --example turn-stress turn.infra.holochain.org 443 test test;
      cargo run --example turn_doctor wss://signal.infra.holochain.org;
    done
    echo success
  )
```

PR #84, opened 2024-03-26
Allow connections to be closed from the public API

ThetaSinner:


crates/tx5-go-pion/src/peer_con.rs
```diff
@@ -95,6 +95,7 @@ impl From<&AnswerConfig> for GoBufRef<'static> {
 
 pub(crate) struct PeerConCore {
     peer_con_id: usize,
+    con_state: PeerConnectionState,
```
ThetaSinner, line 98: Somewhat torn about this. We shouldn't try to close connections that aren't in `connected` or `failed` but I'm not sure we can access this object while the connection is being established. We also have the option to ask for the connection state by calling Go, but that requires an unsafe call to lookup when we've already been told what the state is. Either way, I'm not enforcing this check for now, just warning if we are closing connections we shouldn't be.

crates/tx5/src/ep3/peer.rs
```diff
@@ -58,7 +58,6 @@ pub(crate) struct Peer {
     cmd_task: tokio::task::JoinHandle<()>,
     recv_task: tokio::task::JoinHandle<()>,
     data_task: tokio::task::JoinHandle<()>,
-    #[allow(dead_code)]
```
ThetaSinner, line 61: This is used, don't think the allow is needed

crates/tx5/src/ep3/sig.rs
```diff
@@ -311,7 +311,7 @@ impl Sig {
         }
     }
 
-    pub fn ban(&self, id: Id) {
+    pub fn close(&self, id: Id) {
```
ThetaSinner, line 314: I created a separate `close` but it was identical to `ban` which doesn't actually ban, it's just called to close the connection when banning. Made most sense to me to just merge these

crates/tx5/src/ep3.rs
```diff
@@ -424,13 +424,46 @@ impl Ep3 {
...
+            Some((_, fut)) => {
+                let fut = fut.clone();
+                tokio::task::spawn(async move {
+                    if let Ok(sig) = fut.await {
+                        sig.close(peer_id);
+                    } else {
+                        tracing::warn!(
+                            "Unable to close peer connection: {}",
+                            peer_url
+                        );
```
neonphog, line 453: I feel like this will most often show up because the connection was already closed through some other means... so we'd get: "Unable to close peer connection: Closed". I vote for making this `debug!` level.

neonphog: Oh, it prints the url. That'd be even more confusing. Perhaps it should also print the error.

crates/tx5-go-pion/src/peer_con.rs
```diff
@@ -200,14 +202,50 @@ impl PeerConnection {
...
+                match core.con_state {
+                    PeerConnectionState::Connected
+                    | PeerConnectionState::Failed => {
+                        // Allow the connection to be closed
+                    }
+                    _ => {
+                        tracing::warn!(
+                            "Closing peer connection in invalid state: {:?}",
+                            core.con_state
+                        );
```
neonphog, line 246: Why can they not be closed while being established? In the case where two peers simultaneously send offers to each other, one will establish a peerconnection that gets overridden, so this close will be called on that during normal operation.

crates/tx5/src/ep3/sig.rs
neonphog: Haha, yes this makes more sense :+1:

neonphog (approved):
Couple questions / suggestion. Looks awesome.

crates/tx5-go-pion/src/peer_con.rs
ThetaSinner: I was reading something about not closing connections while they're being established but it looks like that was specific to another implementation in a browser context, so no idea if it applies here. It sounds like it doesn't so I'll remove this code :)

ThetaSinner: I read something about problems with doing so but that was about a different implementation so maybe not relevant here. I'll get rid of this stuff

ThetaSinner: This does mean that `con_state` isn't used yet and `get_con_state` isn't visible outside the crate. I think that's okay for now, it was already an incomplete code path that is now just a little more implemented

PR #91, opened 2024-04-29
sbd integration into tx5

neonphog:
DEPENDS ON: https://github.com/holochain/sbd/pull/1
RESOLVES: https://github.com/holochain/sbd/issues/17

- deletes tx5-signal-srv - we will now be using sbd-server
- updates tx5-signal to be a thin wrapper around sbd-e2e-crypto-client
- adds crate tx5-connection which manages connections initially talking over the signal connection (and a placeholder to eventually upgrade to webrtc connections to be added in a follow-on PR)
- updates tx5 endpoint to work with tx5-connection-s


crates/tx5-connection/src/conn.rs
```diff
@@ -0,0 +1,200 @@
...
+                            // invalid
+                            HandshakeReq(_) | HandshakeRes(_) => break,
+                            Message(msg) => {
+                                if msg_send.send(msg).await.is_err() {
+                                    break;
+                                }
+                            }
+                            _ => (),
+                        }
+                    }
```
neonphog, line 155: Here's the webrtc placeholder. Currently we're just dealing with the `Message` type, but eventually we'll pay attention to offers/answers/ice and try to establish a webrtc connection, switching to it if it is successful.

crates/tx5/src/ep.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub fn get_stats(&self) -> stats::Stats {
+        #[cfg(feature = "backend-go-pion")]
+        let backend = stats::StatsBackend::BackendGoPion;
+        #[cfg(feature = "backend-webrtc-rs")]
+        let backend = stats::StatsBackend::BackendWebrtcRs;
+
+        stats::Stats {
+            backend,
+            peer_url_list: self.get_listening_addresses(),
+            connection_list: Vec::new(),
```
neonphog, line 285: Note, this just returns an empty array for now. I'll flesh out the stats once we have the webrtc connections working.

.github/workflows/stale.yml
ThetaSinner, line 1: Just not useful on this repo?

Cargo.toml
```diff
@@ -15,7 +15,7 @@ resolver = "2"
 [workspace.dependencies]
 app_dirs2 = "2.5.5"
 backtrace = "0.3.69"
-base64 = "0.13.0"
+base64 = "0.22.1"
```
ThetaSinner, line 18: 🥳

.github/workflows/stale.yml
neonphog: exactly. The volume isn't high enough to warrant it, and it doesn't have the permanent label... if we ever need it, we can port the one from holochain over here.

ThetaSinner (approved):
This is large as advertised. I think I'm following most of the changes and they make sense. I'm calling this a 'high level' approval where I think the structure of the changes and the code looks good. If there are logic errors in loops or task management etc then I'm probably not going to catch it in this much code :)

PR #92, opened 2024-05-13
webrtc upgrade for tx5 connections

neonphog:
RESOLVES: https://github.com/holochain/sbd/issues/18

On new connection over the signal server, try to establish a webrtc connection. If successful, switch to transferring data over webrtc, otherwise, we just keep sending messages via the signal server.

### DIFF
--- a/crates/tx5-connection/src/framed.rs
+++ b/crates/tx5-connection/src/framed.rs
@@ -24,6 +24,7 @@ impl FramedConnRecv {
 /// the base connection.
 pub struct FramedConn {
     pub_key: PubKey,
+    weak_conn: Weak<Conn>,
     conn: tokio::sync::Mutex<Arc<Conn>>,
     cmd_send: tokio::sync::mpsc::Sender<Cmd>,
     recv_task: tokio::task::JoinHandle<()>,
@@ -143,6 +144,7 @@ impl FramedConn {
         Ok((
             Self {
                 pub_key,
+                weak_conn: Arc::downgrade(&conn),
                 conn: tokio::sync::Mutex::new(conn),
                 cmd_send,
                 recv_task,
@@ -155,6 +157,15 @@ impl FramedConn {
     /// The pub key of the remote peer this is connected to.
     pub fn pub_key(&self) -> &PubKey {
         &self.pub_key
+    }
+
+    /// Returns `true` if we sucessfully connected over webrtc.
+    pub fn is_using_webrtc(&self) -> bool {
+        if let Some(conn) = self.weak_conn.upgrade() {
+            conn.is_using_webrtc()
+        } else {
+            false
+        }
     }
 
     /// Send a message on the connection.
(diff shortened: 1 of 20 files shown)


crates/tx5-connection/src/framed.rs
```diff
@@ -157,6 +159,24 @@ impl FramedConn {
...
+            false
+        }
+    }
+
+    /// Get connection statistics.
+    pub fn get_stats(&self) -> ConnStats {
+        if let Some(conn) = self.weak_conn.upgrade() {
+            conn.get_stats()
+        } else {
+            ConnStats::default()
```
ThetaSinner, line 176: Return `Option::None` if there are no stats around rather than default?

crates/tx5-connection/src/hub.rs
```diff
@@ -225,4 +258,9 @@ impl Hub {
             Err(Error::other("already connected"))
         }
     }
+
+    /// Set a new webrtc config.
+    pub fn set_webrtc_config(&self, webrtc_config: Vec<u8>) {
+        *self.webrtc_config.lock().unwrap() = webrtc_config;
```
ThetaSinner, line 264: Can this be changed once there are already connections?
Does it just apply to new connections and not existing ones?

Might be useful to have a couple of these details in the doc comment?

crates/tx5-core/src/lib.rs
```diff
@@ -33,6 +33,10 @@ pub type BoxFut<'lt, T> =
 #[derive(Debug, Clone, Copy, serde::Serialize, serde::Deserialize)]
 #[serde(crate = "deps::serde", rename_all = "camelCase")]
 pub struct Tx5InitConfig {
+    /// If true, tracing logs from the backend webrtc library will be included.
+    /// Defaults to `false`.
+    pub tracing_enabled: bool,
```
ThetaSinner, line 38: Nice to be able to control this!

crates/tx5/src/ep.rs
```diff
@@ -302,10 +305,30 @@ impl Endpoint {
...
         let backend = stats::StatsBackend::BackendWebrtcRs;
 
+        let connection_list = self
+            .inner
+            .lock()
+            .unwrap()
+            .peer_map
+            .values()
+            .map(|peer| {
+                let stats = peer.get_stats();
```
ThetaSinner, line 315: I guess this is why it's useful to have default stats?

ThetaSinner (approved):
Looks good!

crates/tx5-connection/src/framed.rs
neonphog: Yeah, on the startup side, i think it makes sense for the stats to be zeroed until we start collecting data. On the shutdown side it's a little weirder for them to jump back to zero, but it makes dealing with the api easier. I vote for keeping it direct for now and we can see if the usage gets confusing.

crates/tx5-connection/src/hub.rs
neonphog: done: https://github.com/holochain/tx5/pull/92/commits/f724b0f018ce04e619bb2b2c381f9191bd8f5158

PR #94, opened 2024-06-06
Ensure SigUrls and PeerUrls are always in a canonical form with even well-known ports listed explicitly

neonphog:


matthme (approved):
Take my approval with a grain of salt as I don't feel that I have enough understanding of the context to assess this PR competently.

crates/tx5/src/url.rs
```diff
@@ -1,192 +1,323 @@
...
-                    }
-
-                    Some(PubKey(Arc::new(dec.try_into().unwrap())))
-                }
-            }
-        },
-    };
-
-    Ok((r.to_string(), pub_key))
+struct UrlCanon {
```
ThetaSinner, line 3: Internal struct so doc comment optional but I think it might be useful to have a brief description here?

ThetaSinner (approved):

crates/tx5/src/url.rs
neonphog:
```suggestion
/// SigUrl and PeerUrl share a lot of common parsing code.
/// This internal struct ensures the base urls are "canonical",
/// always explicitly listing even well-known ports 80 and 443.
struct UrlCanon {
```

steveej (approved):

PR #97, opened 2024-07-08
Use app_dirs2 UserCache directory for storing go lib by default, overridable via env variable

mattyg:
Follow up to #69 

After the changes in #77, the pion go lib was still being written to a temporary directory specified by the package `tempfile`, which got its base path from `std::env::temp_dir`. On android this would return `/data/local/tmp`, which is writable in an adb shell, but not by apps in non-rooted devices.

Now it is written to the UserCache directory specified by `app_dirs2`, or can be overridden via the env variable `TX5_CACHE_DIRECTORY`.

The env variable override lets us set the cache directory to `/data/local/tmp` for running tests on android.

### DIFF
--- a/crates/tx5-core/src/file_check.rs
+++ b/crates/tx5-core/src/file_check.rs
@@ -1,6 +1,8 @@
 //! A couple crates that depend on tx5-core need to be able to write/verify
 //! files on system. Enable this `file_check` feature to provide that ability.
 
+use app_dirs2::AppDirsError;
+
 use crate::{Error, Result};
 
 /// A handle to a verified system file. Keep this instance in memory as
@@ -17,20 +19,17 @@ impl FileCheck {
     }
 }
 
-fn get_user_cache_dir() -> Option<std::path::PathBuf> {
-    match app_dirs2::app_root(
+fn get_user_cache_dir() -> Result<std::path::PathBuf, AppDirsError> {
+    app_dirs2::app_root(
         app_dirs2::AppDataType::UserCache,
         &app_dirs2::AppInfo {
             name: "host.holo.tx5",
             author: "host.holo.tx5",
         },
-    ) {
-        Ok(dir) => Some(dir),
-        _ => None,
-    }
+    )
 }
 
-/// Write a file if needed, verify the file, and return a handle to that file.
+/// Write a temp file if needed, verify the file, and return a handle to that file.
 pub fn file_check(
     file_data: &[u8],
     file_hash: &str,
@@ -38,8 +37,9 @@ pub fn file_check(
     file_name_ext: &str,
 ) -> Result<FileCheck> {
     let file_name = format!("{file_name_prefix}-{file_hash}{file_name_ext}");
+    let tmp_dir = get_user_cache_dir()?;
 
-    let pref_path = get_user_cache_dir().map(|mut d| {
+    let pref_path = tmp_dir.map(|mut d| {
         d.push(&file_name);
         d
     });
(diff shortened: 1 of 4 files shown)


crates/tx5-core/src/file_check.rs
```diff
@@ -17,78 +18,71 @@ impl FileCheck {
...
         &app_dirs2::AppInfo {
             name: "host.holo.tx5",
             author: "host.holo.tx5",
         },
-    ) {
-        Ok(dir) => Some(dir),
-        _ => None,
-    }
+    )
+    .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))
```
mattyg, line 29: Should there be a custom tx5 error type that std::io::Error is passed up into?

It feels wrong to pass an option up and then do nothing if it is None. I think we should always be able to get a cache directory and not being able to do so should error.

neonphog:
```suggestion
    .map_err(|err| std::io::Error::other(err))
```

neonphog: I've just been moving mostly towards using std::io::Error in all my rust code for interoperability. If an api needs to match on a specific variant, you can [downcast](https://doc.rust-lang.org/std/io/struct.Error.html#method.downcast) which admittedly is a little awkward, but not as bad in my opinion as libs all using a million different error types or anyhow or whatever.

crates/tx5-core/src/file_check.rs
```diff
@@ -145,11 +157,14 @@ fn validate(path: &std::path::PathBuf, hash: &str) -> Result<std::fs::File> {
...
-fn write(file_data: &[u8]) -> Result<tempfile::NamedTempFile> {
+/// Write a temp file in the given directory
+fn write(
+    parent_dir: PathBuf,
+    file_data: &[u8],
+) -> Result<tempfile::NamedTempFile> {
     use std::io::Write;
 
-    let mut tmp = tempfile::NamedTempFile::new()?;
+    let mut tmp = tempfile::NamedTempFile::new_in(parent_dir)?;
```
mattyg, line 167: If the file didn't exist in the expected directory, write it to that directory.

crates/tx5-core/src/file_check.rs
```diff
@@ -225,4 +240,48 @@ mod tests {
             let _ = std::fs::remove_file(&path);
         }
     }
+
+    #[test]
+    fn file_check_env_variable_override() {
```
mattyg, line 245: Test that the env variable override works

crates/tx5-core/src/file_check.rs
```diff
@@ -17,78 +18,85 @@ impl FileCheck {
...
-            author: "host.holo.tx5",
-        },
-    ) {
-        Ok(dir) => Some(dir),
-        _ => None,
+/// Get a path to a cache directory where the dependency Pion library can be written
+///
+/// Defaults to the UserCache directory returned by app_dirs2, specific to the target platform.
+/// Overridable via the env variable TX5_CACHE_DIRECTORY
+fn get_cache_dir() -> Result<std::path::PathBuf> {
```
mattyg, line 25: The env variable is only used for running tests on android, to override the directory to `/data/local/tmp`. 

Alternatively we could use `#[cfg(all(test, target_os="android"))]`, but for some reason it wasn't recognizing the target_os as android when running tests in CI. I couldn't figure it out.

neonphog (approved):
Woot, thanks for working through this :+1:

PR #101, opened 2024-07-25
multi-sig tests and fixes

neonphog:
RESOLVES: https://github.com/holochain/holochain/issues/4122

- Adds an integration test to prove out connecting to multiple different signal servers.
- Fixes a bug where we would emit a listening event on outgoing non-listening sbd server connections.

### DIFF
--- a/crates/tx5/tests/tests/multi_sig.rs
+++ b/crates/tx5/tests/tests/multi_sig.rs
@@ -1,0 +1,87 @@
+use std::sync::Arc;
+
+async fn sbd() -> sbd_server::SbdServer {
+    let config = sbd_server::Config {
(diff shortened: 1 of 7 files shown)


crates/tx5/tests/tests/multi_sig.rs
```diff
@@ -0,0 +1,87 @@
...
+                assert_eq!(msg, message);
+                return peer_url;
+            }
+            _ => (),
+        }
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn multi_sig() {
```
ThetaSinner, line 50: Nice!

ThetaSinner (approved):
Looks good!

PR #102, opened 2024-08-02
Fix semaphore dead lock when peer connects to itself

### LINKED ISSUE #143: [BUG] Connection to self dead locks
When connecting to oneself, a send doesn't succeed and keeps the connection in a dead lock.

Reproduction of the problem #102

jost-s:
resolves #143


neonphog (approved):

ThetaSinner (commented):
This is a useful check and I think worth merging. 

I'm wondering though if we're leaving a problem behind here. The "connect to self" was a way to make the connection attempt fail but I thought the deadlock could happen on a failed connection attempt in general?

jost-s:
I don't know if something has changed between when the issue and test were written and now, but there's no dead lock I can produce connecting to self or a non-existing peer.

I've added a test for that anyway.

ThetaSinner (approved):
Nice!

PR #104, opened 2024-08-13
connection buffer and close fixes

neonphog:
- updates sbd-client to include a rate-limit fix
- increases message buffer size on transition to webrtc
- fixes to slow app logic so connections are closed properly

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,8 +45,8 @@ ring = "0.16.20"
 rustls = { version = "0.20.0", features = [ "dangerous_configuration" ] }
 rustls-native-certs = "0.6.2"
 rustls-pemfile = "1.0.2"
-sbd-e2e-crypto-client = "0.0.5-alpha"
-sbd-server = "0.0.5-alpha"
+sbd-e2e-crypto-client = "0.0.6-alpha"
+sbd-server = "0.0.6-alpha"
 serde = { version = "1.0.160", features = [ "derive", "rc" ] }
 serde_json = { version = "1.0.96", features = [ "preserve_order" ] }
 sha2 = "0.10.6"
--- a/crates/tx5-connection/src/conn.rs
+++ b/crates/tx5-connection/src/conn.rs
@@ -1,15 +1,21 @@
 use super::*;
 use std::sync::atomic::Ordering;
 
+/// Message count allowed to accumulate on the boundary switching to webrtc.
+/// This prevents potentially faster to transfer webrtc messages from being
+/// delivered out of order before trailing slower sbd messages are received.
+const MAX_WEBRTC_BUF: usize = 512;
+
 pub(crate) enum ConnCmd {
     SigRecv(tx5_signal::SignalMessage),
     SendMessage(Vec<u8>),
     WebrtcMessage(Vec<u8>),
     WebrtcReady,
+    WebrtcClosed,
 }
 
 /// Receive messages from a tx5 connection.
-pub struct ConnRecv(tokio::sync::mpsc::Receiver<Vec<u8>>);
+pub struct ConnRecv(CloseRecv<Vec<u8>>);
 
 impl ConnRecv {
     /// Receive up to 16KiB of message data.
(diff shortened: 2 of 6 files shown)


crates/tx5-connection/src/conn.rs
```diff
@@ -367,62 +474,110 @@ impl Conn {
...
+                                    DEBUG,
+                                    pub_key = ?pub_key2,
+                                    a = "close: msg_send closed",
+                                );
                                 break;
                             }
                         } else {
                             webrtc_message_buffer.push(msg);
-                            if webrtc_message_buffer.len() > 32 {
+                            if webrtc_message_buffer.len() > MAX_WEBRTC_BUF {
```
neonphog, line 528: this is the buffer fix. 32 just wasn't enough

crates/tx5-connection/src/conn.rs
```diff
@@ -367,62 +474,110 @@ impl Conn {
...
+                        netaudit!(
+                            DEBUG,
+                            pub_key = ?pub_key2,
+                            a = "close: webrtc closed",
+                        );
+                        break;
+                    }
+                }).is_err() {
+                    tracing::warn!(slow_task, "slow app on conn loop task");
+                    break;
```
neonphog, line 580: This is the slow app timeout that will actually get triggered most of the time

crates/tx5-connection/src/webrtc.rs
```diff
@@ -200,15 +213,25 @@ async fn task(
...
             Cmd::BufferedAmountLow => {
+                slow_task = "buf-low";
                 pend_buffer.clear();
             }
-        }
+        }) {
+            Err(_) => {
+                let err = format!("slow app on webrtc loop task: {slow_task}");
+                tracing::warn!("{err}");
+                Err(Error::other(err))
```
neonphog, line 231: This error will now correctly propagate the connection shutdown even if that other slow app timeout is not triggered in conn.rs.

crates/tx5-connection/src/lib.rs
```diff
@@ -66,20 +95,23 @@ impl<T: 'static + Send> Clone for CloseSend<T> {
 impl<T: 'static + Send> Drop for CloseSend<T> {
     fn drop(&mut self) {
         if self.close_on_drop {
-            self.sender.lock().unwrap().take();
+            let s = self.sender.lock().unwrap().take();
+            if let Some(mut s) = s {
+                s.close_channel();
```
neonphog, line 100: This switch to a futures::channel instead of a tokio lets us explicitly close the channel from the sender side. It turned out this fix is not technically needed, but I left it in because it's more explicit about our intentions here.

Cargo.toml
```diff
@@ -45,8 +45,8 @@ ring = "0.16.20"
 rustls = { version = "0.20.0", features = [ "dangerous_configuration" ] }
 rustls-native-certs = "0.6.2"
 rustls-pemfile = "1.0.2"
-sbd-e2e-crypto-client = "0.0.5-alpha"
-sbd-server = "0.0.5-alpha"
+sbd-e2e-crypto-client = "0.0.6-alpha"
```
neonphog, line 48: upgrading the sbd-client contains a rate-limiting fix.

crates/tx5-connection/src/lib.rs
```diff
@@ -33,6 +33,26 @@ compile_error!("Must specify exactly 1 webrtc backend");
 #[cfg(feature = "backend-go-pion")]
 pub use tx5_go_pion::Tx5InitConfig;
 
+/// Grace time to allow a slow app to catch up before we close a
+/// connection to prevent our memory from filling up with backlogged
+/// message data.
+const SLOW_APP_TO: std::time::Duration = std::time::Duration::from_millis(99);
```
jost-s, line 39: Why 99 ms?

jost-s (approved):

ThetaSinner (approved):

crates/tx5-connection/src/lib.rs
neonphog: Originally it was 19ms to keep it under 20ms for tokio, but this version is compatible with async code so it can be a bit longer for some more leeway.

PR #105, opened 2024-10-09
abstract tx5 backend in prep for m̶o̶c̶k̶  mem backend

neonphog:
- Actual mock backend PR to follow.
- There are a couple apis that aren't great encapsulation for the abstraction at this level, but in the name of expediency, I've just called them out in the code for now to be addressed at a later date.
- I'd love for the mock config to be configurable all the way up into the holochain conductor config, so I went with a generic `serde_json::Value` because those items will not be relevant to any other backends, but we need a shared abstraction. I'm happy to have feedback / suggestions of different ways to do this.

### DIFF
--- a/crates/tx5/src/backend.rs
+++ b/crates/tx5/src/backend.rs
@@ -1,0 +1,134 @@
+//! Backend modules usable by tx5.
+
+use std::io::Result;
+use std::sync::Arc;
+
+use futures::future::BoxFuture;
+
+use crate::{Config, PubKey};
+use tx5_core::deps::serde_json;
+
+#[cfg(feature = "backend-go-pion")]
+mod go_pion;
+
+/// Backend modules usable by tx5.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum BackendModule {
+    #[cfg(feature = "backend-go-pion")]
+    /// The Go Pion-based backend.
+    GoPion,
+
+    #[cfg(feature = "backend-webrtc-rs")]
+    /// The Webrtc-RS-based backend.
+    WebrtcRs,
+
+    /// The mock backend.
+    Mock,
+}
+
+impl Default for BackendModule {
+    #[allow(unreachable_code)]
+    fn default() -> Self {
+        #[cfg(feature = "backend-go-pion")]
+        return Self::GoPion;
+        #[cfg(feature = "backend-webrtc-rs")]
+        return Self::WebrtcRs;
+        Self::Mock
+    }
+}
+
+impl BackendModule {
+    /// Get a default version of the module-specific config.
+    pub fn default_config(&self) -> serde_json::Value {
+        match self {
+            #[cfg(feature = "backend-go-pion")]
+            Self::GoPion => go_pion::default_config(),
+            #[cfg(feature = "backend-webrtc-rs")]
+            Self::WebrtcRs => todo!(),
+            Self::Mock => serde_json::json!({}),
+        }
+    }
+
+    /// Connect a new backend module endpoint.
+    pub async fn connect(
+        &self,
+        url: &str,
+        listener: bool,
+        config: &Arc<Config>,
+    ) -> Result<(DynBackEp, DynBackEpRecv)> {
+        match self {
+            #[cfg(feature = "backend-go-pion")]
+            Self::GoPion => go_pion::connect(config, url, listener).await,
+            #[cfg(feature = "backend-webrtc-rs")]
+            Self::WebrtcRs => todo!(),
+            Self::Mock => todo!(),
+        }
+    }
+}
+
+/// Backend connection.
+pub trait BackCon: 'static + Send + Sync {
+    /// Send data over this backend connection.
+    fn send(&self, data: Vec<u8>) -> BoxFuture<'_, Result<()>>;
+
+    /// Get the pub_key identifying this connection.
+    fn pub_key(&self) -> &PubKey;
+
+    /// Returns `true` if we successfully connected over webrtc.
+    // TODO - this isn't good encapsulation
+    fn is_using_webrtc(&self) -> bool;
+
+    /// Get connection statistics.
+    // TODO - this isn't good encapsulation
+    fn get_stats(&self) -> tx5_connection::ConnStats;
+}
+
+/// Trait-object version of backend connection.
+pub type DynBackCon = Arc<dyn BackCon + 'static + Send + Sync>;
+
+/// Backend connection receiver.
+pub trait BackConRecv: 'static + Send {
+    /// Receive data from this backend connection.
+    fn recv(&mut self) -> BoxFuture<'_, Option<Vec<u8>>>;
+}
+
+/// Trait-object version of backend connection receiver.
+pub type DynBackConRecv = Box<dyn BackConRecv + 'static + Send>;
+
+/// Pending connection.
+pub trait BackWaitCon: 'static + Send {
+    /// Wait for the connection
+    fn wait(
+        &mut self,
+        // TODO - this isn't good encapsulation
+        recv_limit: Arc<tokio::sync::Semaphore>,
+    ) -> BoxFuture<'static, Result<(DynBackCon, DynBackConRecv)>>;
+
+    /// Get the pub_key identifying this connection.
+    fn pub_key(&self) -> &PubKey;
+}
+
+/// Trait-object version of backend wait con.
+pub type DynBackWaitCon = Box<dyn BackWaitCon + 'static + Send>;
+
+/// Backend endpoint.
+pub trait BackEp: 'static + Send + Sync {
+    /// Establish an outgoing connection from this backend endpoint.
+    fn connect(&self, pub_key: PubKey)
+        -> BoxFuture<'_, Result<DynBackWaitCon>>;
+
+    /// Get the pub_key identifying this endpoint.
+    fn pub_key(&self) -> &PubKey;
+}
+
+/// Trait-object version of backend endpoint.
+pub type DynBackEp = Arc<dyn BackEp + 'static + Send + Sync>;
+
+/// Backend endpoint receiver.
+pub trait BackEpRecv: 'static + Send {
+    /// Receive incoming connection from this backend endpoint.
+    fn recv(&mut self) -> BoxFuture<'_, Option<DynBackWaitCon>>;
+}
+
+/// Trait-object version of backend endpoint receiver.
+pub type DynBackEpRecv = Box<dyn BackEpRecv + 'static + Send>;
--- a/crates/tx5/src/backend.rs
+++ b/crates/tx5/src/backend.rs
@@ -55,7 +55,7 @@ impl BackendModule {
         url: &str,
         listener: bool,
         config: &Arc<Config>,
-    ) -> Result<(DynBackEp, DynBackEpRecv)> {
+    ) -> Result<(DynBackEp, DynBackEpRecvCon)> {
         match self {
             #[cfg(feature = "backend-go-pion")]
             Self::GoPion => go_pion::connect(config, url, listener).await,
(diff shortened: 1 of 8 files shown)


crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
...
+mod go_pion;
+
+/// Backend modules usable by tx5.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum BackendModule {
+    #[cfg(feature = "backend-go-pion")]
+    /// The Go Pion-based backend.
+    GoPion,
+
+    #[cfg(feature = "backend-webrtc-rs")]
```
ThetaSinner, line 21: How is this doing? Is it coming along at all or still not production ready?

crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
...
+    WebrtcRs,
+
+    /// The mock backend.
+    Mock,
+}
+
+impl Default for BackendModule {
+    #[allow(unreachable_code)]
+    fn default() -> Self {
+        #[cfg(feature = "backend-go-pion")]
```
ThetaSinner, line 32: Are these already prevented from being enabled together?

crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
...
+    /// Get the pub_key identifying this connection.
+    fn pub_key(&self) -> &PubKey;
+
+    /// Returns `true` if we successfully connected over webrtc.
+    // TODO - this isn't good encapsulation
+    fn is_using_webrtc(&self) -> bool;
+
+    /// Get connection statistics.
+    // TODO - this isn't good encapsulation
+    fn get_stats(&self) -> tx5_connection::ConnStats;
```
ThetaSinner, line 83: I would like it if we got a structure back based on the currently enabled back end. I was fine with tx5 vs tx2 just returning different data and I think the same is still true with go-pion vs webrtc-rs vs mock. Even if it is different based on the enabled back end, that's what I'd like to see presented up at the conductor level.

Maybe an enum could work and then the encapsulation is reasonable

crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
...
+pub trait BackCon: 'static + Send + Sync {
+    /// Send data over this backend connection.
+    fn send(&self, data: Vec<u8>) -> BoxFuture<'_, Result<()>>;
+
+    /// Get the pub_key identifying this connection.
+    fn pub_key(&self) -> &PubKey;
+
+    /// Returns `true` if we successfully connected over webrtc.
+    // TODO - this isn't good encapsulation
+    fn is_using_webrtc(&self) -> bool;
```
ThetaSinner, line 79: What would the right interface look like here? Something like an `inner` that let's you get at the real connection and you have to know which ones it makes sense to ask about the use of webrtc as the caller?

crates/tx5/src/backend/go_pion.rs
```diff
@@ -0,0 +1,136 @@
...
+
+struct GoConRecv(tx5_connection::FramedConnRecv);
+
+impl BackConRecv for GoConRecv {
+    fn recv(&mut self) -> BoxFuture<'_, Option<Vec<u8>>> {
+        Box::pin(async { self.0.recv().await })
+    }
+}
+
+struct GoWaitCon(
```
ThetaSinner, line 34: When the inner fields aren't trivial, I'd prefer named fields

ThetaSinner: Although I see these get extracted into named fields right away in the function below... So I guess as long as the code is written that way, you don't actually have to work too hard to work out what the `0`, `1` and `2` are

crates/tx5/src/config.rs
```diff
@@ -81,6 +91,8 @@ impl Default for Config {
             backoff_start: std::time::Duration::from_secs(5),
             backoff_max: std::time::Duration::from_secs(60),
             preflight: None,
+            backend_module: crate::backend::BackendModule::default(),
```
ThetaSinner, line 94: Should this only be compile time configurable? Just thinking about writing tests against this, we'd have to build and test one way to run most tests and then build and test again with different features to use mocking

crates/tx5/src/backend.rs
neonphog: I wish I knew, but it doesn't seem to be the appropriate time to evaluate it. Once we have everything stable it'd be great to just go straight for the implementation so we can do a direct A/B comparison through our CI and manual usage.

crates/tx5/src/backend.rs
neonphog: They are currently, but when we get to implementing it, I'd actually like them to *not* be exclusive, hence the fallback pattern in this default constructor. Not that we'd compile very often with both of them, but it would fit better into rust's features being additive paradigm.

crates/tx5/src/backend.rs
neonphog: For sure that could work. Alternately, we could have this return serde_json::Value and have concrete types exposed that can be transcoded into if you know the backend you're using.

For now, I was just going to use this and the mock backend will return a blank ConnStats...

crates/tx5/src/backend.rs
neonphog: It's the latter. "Connections" can communicate directly through sbd, then they attempt to upgrade to a webrtc connection. So `is_using_webrtc()` returns true once/if that upgrade is successful.

crates/tx5/src/backend/go_pion.rs
neonphog: Yeah, that's fair. It started out as one, then gradually went up to three. I should switch it to the struct form.

crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
...
+impl Default for BackendModule {
+    #[allow(unreachable_code)]
+    fn default() -> Self {
+        #[cfg(feature = "backend-go-pion")]
+        return Self::GoPion;
+        #[cfg(feature = "backend-webrtc-rs")]
+        return Self::WebrtcRs;
+        Self::Mock
+    }
+}
```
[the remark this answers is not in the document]
neonphog: My hope was to make the backend features non-exclusive, hence the fallback returns.

crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
...
+    // TODO - this isn't good encapsulation
+    fn is_using_webrtc(&self) -> bool;
+
+    /// Get connection statistics.
+    // TODO - this isn't good encapsulation
+    fn get_stats(&self) -> tx5_connection::ConnStats;
+}
+
+/// Trait-object version of backend connection.
+pub type DynBackCon = Arc<dyn BackCon + 'static + Send + Sync>;
```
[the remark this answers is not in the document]
neonphog: Type-erasure. The associated types (or sub-generics) quickly get ridiculous with the stack of traits like this.

crates/tx5/src/config.rs
neonphog: Your question seems to answer why I wouldn't want to switch this to compile-time configuration. This way, most of the tests can use the true backend, but I can also have some which exercise the mock without having to have multiple top-level calls to `cargo test` with different features.

crates/tx5/src/backend.rs
neonphog: > What would the right interface look like here?

Yeah, that's still TBD. Does it even make sense to have this call be separate? Maybe it should just be part of the stats call, although that would just defer the problem, because there's an encapsulation question about that too : )

crates/tx5/src/backend/go_pion.rs
```diff
@@ -0,0 +1,136 @@
...
+    ) -> BoxFuture<'static, Result<(DynBackCon, DynBackConRecv)>> {
+        let con = self.1.take();
+        let con_recv = self.2.take();
+        Box::pin(async move {
+            let (con, con_recv) = match (con, con_recv) {
+                (_, None) | (None, _) => {
+                    return Err(std::io::Error::other("already awaited"))
+                }
+                (Some(con), Some(con_recv)) => (con, con_recv),
+            };
```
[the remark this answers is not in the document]
neonphog: Indeed. I'll decline the functional change, but I was able to compact the match a bit by flipping the order and using a catch all: https://github.com/holochain/tx5/pull/105/commits/681fd982b431bff9300833f093e9a472f91e3dbd

crates/tx5/src/backend/go_pion.rs
neonphog: done: https://github.com/holochain/tx5/pull/105/commits/681fd982b431bff9300833f093e9a472f91e3dbd

crates/tx5/src/backend.rs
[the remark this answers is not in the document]
jost-s: What happens in Callums suggestion when you compile with both features? Compiler error?

crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
...
+/// Trait-object version of backend endpoint.
+pub type DynBackEp = Arc<dyn BackEp + 'static + Send + Sync>;
+
+/// Backend endpoint receiver.
+pub trait BackEpRecv: 'static + Send {
+    /// Receive incoming connection from this backend endpoint.
+    fn recv(&mut self) -> BoxFuture<'_, Option<DynBackWaitCon>>;
+}
+
+/// Trait-object version of backend endpoint receiver.
```
jost-s, line 133: What's the difference between this and the backend connection receiver?

crates/tx5/src/backend.rs
```diff
@@ -0,0 +1,134 @@
+//! Backend modules usable by tx5.
+
+use std::io::Result;
+use std::sync::Arc;
+
+use futures::future::BoxFuture;
+
+use crate::{Config, PubKey};
+use tx5_core::deps::serde_json;
```
jost-s, line 9: Is rmp_serde an alternative worth considering?

crates/tx5/src/backend.rs
jost-s: Maybe that's a different topic, but why does tx5 need to know whether it's connected indirectly through sbd or directly through webrtc?

crates/tx5/src/backend.rs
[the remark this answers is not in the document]
cdunster: Yeah it would be a compiler error for multiple defined defaults. You could fix it with more `cfg_attr` like setting `WebrtcRs` to only default if `go-pion` isn't enabled:

```rust
#[cfg(feature = "backend-webrtc-rs")]
#[cfg_attr(not(feature = "backend-go-pion"), default)]
/// The Webrtc-RS-based backend.
WebrtcRs,
```

but I can imagine that this starts to look unruly as we possibly add more backends.

crates/tx5/src/backend.rs
neonphog: As far as I remember, it's informational or for unit tests. But I still submit this is something that doesn't have to be sorted out to merge this PR, since I'm not changing any functionality here.

crates/tx5/src/backend.rs
neonphog: The endpoint receiver receives incoming connections. The connection receiver receives incoming messages over that one connection.

crates/tx5/src/backend.rs
neonphog: We need to have serde_json as a dependency anyways, because the webrtc signaling messages are in json. In *this* module I'm using it for the generic `Value` type. There are other generic type crates in the rust ecosystem, but if we have to have this one anyways, we should just use it.

crates/tx5/src/backend.rs
jost-s: I asked out of curiosity, it wasn't meant as a subtle push to change it. I prefer being explicit about pushes ;-) Fine by me to keep as is.

jost-s (approved):

crates/tx5/src/backend.rs
jost-s: Now that I look again at the trait functions it's pretty clear. I guess the trait names could speak more to that, e. g. `ConnRecv` and `DataRecv`

neonphog: I'd personally find that confusing, since then you'd be given a `ConnRecv` when creating an endpoint, and not when opening a connection...

jost-s: Right, maybe `EpConnRecv` would be better. I find more confusing that a backend connection receiver does not receive connections, but messages/data. =D

neonphog: Fair enough: https://github.com/holochain/tx5/pull/105/commits/885af263ffc4b31fb27e37c9a67d74f983df79c1

Went with:
- `EpRecvCon`
- `ConRecvData`

cdunster (approved):

PR #106, opened 2024-10-10
mem backend for tx5

neonphog:


maackle:
sweet, perhaps with this we can use this to finish up the [tx2 removal in holochain](https://github.com/holochain/holochain/pull/3225)?

crates/tx5/examples/mem-echo-stress.rs
```diff
@@ -0,0 +1,94 @@
...
+/*
+ * Mem backend timings. The hope is to have near ideal zero overhead.
+ * At time of writing, you can see we're pretty close to exactly
+ * proportional to node count for a given machine's resources:
+ *
+ * - 1 nodes, 33941 ops, 0.000028 sec/op, in 1.004478s
+ * - 10 nodes, 174041 ops, 0.000056 sec/op, in 1.016977s
+ * - 100 nodes, 175111 ops, 0.000580 sec/op, in 1.018943s
+ * - 1000 nodes, 160266 ops, 0.006361 sec/op, in 1.019859s
+ */
```
cdunster, line 13: Shouldn't this be a file-level comment block with `//!` at the top?

crates/tx5/src/backend/mem.rs
```diff
@@ -0,0 +1,333 @@
...
+
+impl BackEpRecvCon for EpRecvCon {
+    fn recv(&mut self) -> BoxFuture<'_, Option<DynBackWaitCon>> {
+        Box::pin(async { self.0.recv().await })
+    }
+}
+
+fn gen_pub_key(uniq: u64, loc: usize) -> PubKey {
+    // in base64 a string of 0xb7 renders as `t7e3t7e3t...`
+    // which sort-of looks like "test" if you squint : )
```
cdunster, line 104: :laughing:

crates/tx5/src/backend/mem.rs
```diff
@@ -0,0 +1,333 @@
...
+    let mut pub_key = [0xb7; 32];
+    pub_key[16..24].copy_from_slice(&uniq.to_be_bytes());
+    pub_key[24..32].copy_from_slice(&(loc as u64).to_be_bytes());
+
+    PubKey(Arc::new(pub_key))
+}
+
+fn parse_pub_key(pub_key: &PubKey) -> (u64, usize) {
+    let mut uniq = [0; 8];
+    uniq.copy_from_slice(&(*pub_key)[16..24]);
```
cdunster, line 114: Is the `*` dereference necessary? Can't we just do:

```suggestion
    uniq.copy_from_slice(&pub_key[16..24]);
```

crates/tx5/src/backend/mem.rs
```diff
@@ -0,0 +1,333 @@
...
+    let uniq = u64::from_be_bytes(uniq);
+
+    let mut loc = [0; 8];
+    loc.copy_from_slice(&(*pub_key)[24..32]);
```
cdunster, line 118: Same here:

```suggestion
    loc.copy_from_slice(&pub_key[24..32]);
```

cdunster (commented):
Some minor comments

crates/tx5/examples/mem-echo-stress.rs
neonphog: ah, good call. done: https://github.com/holochain/tx5/pull/106/commits/1887abd29272df5fff010ef739bfee5176ec4289

crates/tx5/examples/mem-echo-stress.rs
```diff
@@ -0,0 +1,92 @@
+//! Mem backend timings. The hope is to have near ideal zero overhead.
+//! At time of writing, you can see we're pretty close to exactly
```
jost-s, line 2: What is proportional? 1 node, 33941 ops, then 10 x the number of nodes and 174041 ops which is about 5 x as many ops. Then again 10 x as many nodes with 175111 ops, roughly the same.

Regarding time/op it doubles from 1 to 10 nodes, then rises 10 fold as does the number of nodes, but that's the only proportionality I can make out.

crates/tx5/examples/mem-echo-stress.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    tokio::task::spawn(async move {
+        while let Some(evt) = listen_recv.recv().await {
+            match evt {
+                EndpointEvent::ListeningAddressOpen { .. } => (),
+                EndpointEvent::Connected { .. } => (),
+                EndpointEvent::Message { peer_url, .. } => {
+                    let listen_ep = listen_ep.clone();
+                    // Don't hold up our recv loop on the response
+                    tokio::task::spawn(async move {
```
jost-s, line 34: Is it faster if you create just one task for responses with a channel that sends the response when it receives something from the trigger, and just trigger that channel for every incoming message?

cdunster (approved):

crates/tx5/examples/mem-echo-stress.rs
neonphog: @jost-s - you got it exactly. 1 node is different because it hadn't yet maxed out my cpu usage. Once there are 10 nodes, my cpu is maxed and the time/op increases linearly with the number of nodes plus a little overhead.

crates/tx5/examples/mem-echo-stress.rs
neonphog: It might be, but this is overhead in the test setup, and I was happy with the overall overhead, so don't feel the need to optimize further.

PR #107, opened 2024-10-11
Bump Lair keystore to 0.4.6

ThetaSinner:
Update Lair Keystore API and bump tx5 to release

### DIFF
--- a/crates/tx5-go-pion-sys/Cargo.toml
+++ b/crates/tx5-go-pion-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tx5-go-pion-sys"
-version = "0.0.14-alpha"
+version = "0.0.15-alpha"
 edition = "2021"
 description = "Low level rust bindings to the go pion webrtc library"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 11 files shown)


crates/tx5-go-pion-sys/Cargo.toml
```diff
@@ -47,3 +47,6 @@ force_dynamic_link = []
 # forces the code to run through the static linking process
 # even if that may not be supported by the target platform
 force_static_link = []
+
+[lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(link_dynamic)', 'cfg(link_static)'] }
```
neonphog, line 52: thanks!

neonphog (approved):
woot

crates/tx5-go-pion-sys/Cargo.toml
ThetaSinner: No worries, already ran into this same funky clippy behavior on Holochain so it's an easy fix

From what I've read, they're going to give us cleaner ways of expressing custom cfgs at some point

Issue #110, opened 2024-10-16
re-evaluate libdatachannel as a backend


neonphog:
https://crates.io/crates/datachannel

since it is now [used as the WHIP client for obs-studio](https://github.com/obsproject/obs-studio/pull/7926)

PR #111, opened 2024-10-17
libdatachannel backend impl

neonphog:
RESOLVES https://github.com/holochain/tx5/issues/110

- Minor refactor of tx5-connection logic to enable swap-able webrtc backends while keeping the sbd signaling logic.
- Addition of libdatachannel as a webrtc backend
- Unfortunately, libdatachannel isn't immediately ready for production, so, keeping go-pion as the default for now, because:
  - windows ci has a flaky failure to connect over webrtc when libdatachannel backend is used
  - libdatachannel fails to build for the android target because it's not set up well to cross-compile

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,8 +17,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [
-          #ubuntu-latest,
-          #macos-latest,
+          ubuntu-latest,
+          macos-latest,
           windows-latest,
         ]
         toolchain: [
@@ -64,12 +64,11 @@ jobs:
           RUST_BACKTRACE: 1
         run: |-
           $env:SODIUM_LIB_DIR="$(pwd)\vcpkg\packages\libsodium_x64-windows-release\lib"
-          #make unit
-          cargo test --manifest-path crates/tx5/Cargo.toml webrtc_transition_ordering -- --nocapture
+          make unit
 
       - name: Cargo Test
         if: matrix.os != 'windows-latest'
         env:
-          RUST_LOG: warn
+          RUST_LOG: trace
           RUST_BACKTRACE: 1
         run: make unit
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,16 +51,6 @@ jobs:
           manifest-dir: ${{ github.workspace }}/.github/manifest
           github-binarycache: true
 
-      - name: Cargo Build Windows
-        if: matrix.os == 'windows-latest'
-        run: |-
-          $env:SODIUM_LIB_DIR="$(pwd)\vcpkg\packages\libsodium_x64-windows-release\lib"
-          cargo build --all-targets
-
-      - name: Cargo Build
-        if: matrix.os != 'windows-latest'
-        run: cargo build --all-targets
-
       - name: Cargo Build - musl
         if: matrix.os == 'ubuntu-latest'
         run: |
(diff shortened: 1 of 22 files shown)


Makefile
```diff
@@ -77,13 +77,17 @@ test: static unit
 
 unit:
 	cargo build --all-targets
-	RUST_BACKTRACE=1 RUST_LOG=error cargo test -- --nocapture
+	RUST_BACKTRACE=1 RUST_LOG=info cargo test -- --nocapture
+	#--TODO--Once libdatachannel is the default, we'll want to keep
+	#        go-pion tested for a while until we're ready to deprecate it
+	#RUST_BACKTRACE=1 RUST_LOG=info cargo test --no-default-features --features backend-go-pion --manifest-path crates/tx5-connection/Cargo.toml -- --nocapture
```
jost-s, line 83: Can you not run tests for the entire workspace?

.github/workflows/test.yml
```diff
@@ -51,16 +51,6 @@ jobs:
...
-      - name: Cargo Build Windows
-        if: matrix.os == 'windows-latest'
-        run: |-
-          $env:SODIUM_LIB_DIR="$(pwd)\vcpkg\packages\libsodium_x64-windows-release\lib"
-          cargo build --all-targets
-
-      - name: Cargo Build
-        if: matrix.os != 'windows-latest'
-        run: cargo build --all-targets
-
```
jost-s, line 63: Different builds no longer required, I suppose?

crates/tx5-connection/src/webrtc/go_pion.rs
jost-s, line 1: this is just moved from webrtc, I take ti?

crates/tx5-connection/src/webrtc/libdatachannel.rs
```diff
@@ -0,0 +1,382 @@
...
+    send_buffer: usize,
+    mut evt_send: CloseSend<WebrtcEvt>,
+    cmd_send: CloseSend<Cmd>,
+    mut cmd_recv: CloseRecv<Cmd>,
+) -> Result<()> {
+    evt_send.set_close_on_drop(true);
+
+    #[derive(Debug, Default, serde::Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    struct U {
```
jost-s, line 226: U for something that contains urls, and C for something that contains ice servers? connections?

Makefile
neonphog: The line that remains is testing the entire workspace with the default feature.
The tests I commented out represent the 2 crates that are actually affected by having a different backend feature, so there is no need to run the other crates again.

.github/workflows/test.yml
neonphog: cargo build has been moved into the Makefile, so the `make unit` below now covers this step.

crates/tx5-connection/src/webrtc/go_pion.rs
neonphog: correct :+1:

crates/tx5-connection/src/webrtc/libdatachannel.rs
neonphog: Hehe, something like that. Since these structs are only used inline, and not outside this function I wasn't too concerned about the names, but if you think it's confusing I can try to come up with something : )

jost-s: Ah, fine by me as it is.

jost-s (approved):

ThetaSinner (approved):

PR #114, opened 2024-11-11
Switch to add_parsable_certificates

matthme:
Should address https://github.com/holochain/holochain/issues/4418

### DIFF
--- a/crates/tx5-signal/Cargo.toml
+++ b/crates/tx5-signal/Cargo.toml
@@ -31,8 +31,6 @@ tokio-tungstenite = { workspace = true }
 tracing = { workspace = true }
 tx5-core = { workspace = true }
 url = { workspace = true }
-
-[target.'cfg(not(any(target_os = "windows", target_os = "linux", target_os = "macos")))'.dependencies]
 webpki-roots = { workspace = true }
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


matthme:
According to the documentation the [add](https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.add) method is apparently known to fail for some certificates so I switched to the [add_parsable_certificates](https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.add_parsable_certificates) method instead now.

crates/tx5-signal/Cargo.toml
```diff
@@ -31,8 +31,6 @@ tokio-tungstenite = { workspace = true }
 tracing = { workspace = true }
 tx5-core = { workspace = true }
 url = { workspace = true }
-
-[target.'cfg(not(any(target_os = "windows", target_os = "linux", target_os = "macos")))'.dependencies]
```
ThetaSinner, line 35: This doesn't appear to be platform specific in any way, but it looks like it was deliberately only included for platforms that won't have a certificate store. Now that we're successfully using the certificate store on windows and this isn't being used on the platforms it's been included for, should this be reverted?

ThetaSinner:
Tested on Windows where this was crashing for me, confirmed that this solves the problem and I can still make connections

crates/tx5-signal/Cargo.toml
neonphog: Yes, this should be reverted. `rustls_native_certs` doesn't currently work on android or ios. (although it'd be nice if it did : )

matthme: Oh yes, sorry. I added this with the initial approach where the certs would be bundled.

neonphog (approved):
:+1:

ThetaSinner (approved):

Issue #112, opened 2024-11-01
Replace "slow app timeout" with bounded channels


neonphog:
go pion and libdatachannel both emit events synchronously. To move into the async world without potentially filling up memory with messages we had implemented "slow app timeouts" in the go integration, but also with some tokio timeout wrappers on app-side handling functions.

A simpler solution that is more resilient to momentary cpu pauses would be to just use a bounded channel and close the connection if we cannot send to the channel. This has been implemented in the experimental libdatachannel backend.

This should also be implemented for the go pion integration, and the timeout wrappers in app logic should be removed.

PR #116, opened 2025-01-14
remove the app timeout

neonphog:
RESOLVES: https://github.com/holochain/tx5/issues/112


jost-s (approved):
LGTM
I read this as "whenever a send of almost any kind fails, close the connection".

I guess that it's up to the tx5 consumer to re-open the connection when it's closed due to inactivity.

crates/tx5-connection/src/webrtc/libdatachannel.rs
```diff
@@ -25,27 +25,29 @@ struct Dch(CloseSend<Cmd>);
 
 impl datachannel::DataChannelHandler for Dch {
     fn on_open(&mut self) {
-        self.0.send_or_close(Cmd::DataChanOpen);
+        let _ = self.0.send_or_close(Cmd::DataChanOpen);
```
ThetaSinner, line 28: No need to report error here because the channel can't be full when the connection is first opened?

crates/tx5-connection/src/webrtc/libdatachannel.rs
```diff
@@ -25,27 +25,29 @@ struct Dch(CloseSend<Cmd>);
...
+        let _ =
+            self.0
+                .send_or_close(Cmd::Error(std::io::Error::other(format!(
+                    "DataChanError: {err}"
+                ))));
     }
 
     fn on_message(&mut self, msg: &[u8]) {
-        self.0.send_or_close(Cmd::RecvMessage(msg.to_vec()));
+        let _ = self.0.send_or_close(Cmd::RecvMessage(msg.to_vec()));
```
ThetaSinner, line 46: Worth logging the error here?

ThetaSinner (approved):
Looks good, couple of little questions about whether there's useful information to be reported from errors

crates/tx5-connection/src/webrtc/libdatachannel.rs
neonphog: I waffled back and forth on both of these, but ultimately, the only error we get is that the channel was closed, which might be because of a filled channel buffer, but equally could just be because the app requested the connection close. All events emitted by the underlying webrtc backend would emit errors that confuse debuggers. I'm going to leave this not emitting for now, but we can always retool this flow if we run into something hard to debug.

PR #122, opened 2025-02-03
test: add failing ep stress test

jost-s:


jost-s:
`ep` is dropped prematurely which leads to the closed connection. No other issues.

Issue #123, opened 2025-02-05
[bug] sending too fast closes connections


neonphog:
with errors such as:

```
close: webrtc buffer overflow
send_framed_error
channel is full
```

Write 2 stress tests, one that sends small messages quickly and one that sends large messages quickly, repair any paths that lead to connections closing by using async backpressure.

PR #124, opened 2025-02-05
add stress tests and fix broken pipe when send too fast

neonphog:
RESOLVES: https://github.com/holochain/tx5/issues/123

### DIFF
--- a/crates/tx5-connection/src/lib.rs
+++ b/crates/tx5-connection/src/lib.rs
@@ -105,6 +105,40 @@ impl<T: 'static + Send> CloseSend<T> {
             Err(ErrorKind::BrokenPipe.into())
         }
     }
+
+    pub async fn send(&self, t: T) -> Result<()> {
+        let res = tokio::time::timeout(
+            // hard-coded to 1 minute for now. This indicates a system
+            // is very backed up, and is here just to prevent forever hangs
+            std::time::Duration::from_secs(60),
+            async {
+                let sender = self.sender.lock().unwrap().clone();
+                if let Some(mut sender) = sender {
+                    use futures::sink::SinkExt;
+                    if sender.send(t).await.is_ok() {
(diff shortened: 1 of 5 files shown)


crates/tx5-connection/src/lib.rs
```diff
@@ -105,6 +105,40 @@ impl<T: 'static + Send> CloseSend<T> {
             Err(ErrorKind::BrokenPipe.into())
         }
     }
+
+    pub async fn send(&self, t: T) -> Result<()> {
+        let res = tokio::time::timeout(
+            // hard-coded to 1 minute for now. This indicates a system
+            // is very backed up, and is here just to prevent forever hangs
+            std::time::Duration::from_secs(60),
```
neonphog, line 113: lemme know if y'all are okay with this as an outside safeguard, or if it needs to be configurable

jost-s (commented):
This begs the question if there's something wrong with the channel, so to say? If the async method does not overflow it but the sync one does.

Is it a problem now that the channel doesn't get closed anymore if sending fails?

neonphog:
> This begs the question if there's something wrong with the channel, so to say? If the async method does not overflow it but the sync one does.

This is exactly the solution... because the async one will wait until there is space before pushing the new message on.

> Is it a problem now that the channel doesn't get closed anymore if sending fails?

The channel does get closed on send failure. There's a match at the end of that function that does it.

jost-s:
> This is exactly the solution... because the async one will wait until there is space before pushing the new message on.

jost-s (approved):

ThetaSinner (approved):
Yay for stress tests and this fix looks reasonable to me. 

I think having some timeout is reasonable for now. It might give more context to require people making a network request to decide their own timeout but that's an adjustment we can make later if needed. It's good to have something in place for now

PR #127, opened 2025-02-10
BREAKING: only use sbd messaging as a fallback for webrtc failure

neonphog:
This is a BREAKING protocol change because if you mix old an new peers, the old peers will send sbd messages that will immediately cause new peers to break out of webrtc mode.

- Switch default backend to libdatachannel
- Before sending any messages, attempt first to connect to webrtc. If unsuccessful, fall back to messaging over sbd.


crates/tx5/Cargo.toml
```diff
@@ -11,7 +11,7 @@ keywords = ["holochain", "holo", "p2p", "webrtc", "networking"]
 categories = ["network-programming"]
 
 [features]
-default = [ "backend-go-pion" ]
+default = [ "backend-libdatachannel" ]
```
neonphog, line 14: WARNING: I'm sneaking in a change to default to libdatachannel here, since we'll have to bump the version to 0.3 anyways! : )

crates/tx5-connection/src/test.rs
```diff
@@ -37,22 +47,17 @@ impl TestSrv {
...
-                .to_vec(),
+                b"{}".to_vec(),
                 &format!("ws://{addr}"),
                 Arc::new(HubConfig {
                     backend_module: BackendModule::default(),
                     signal_config: Arc::new(tx5_signal::SignalConfig {
                         listener: true,
                         allow_plain_text: true,
-                        max_idle: std::time::Duration::from_secs(1),
+                        max_idle: std::time::Duration::from_secs(max_idle_secs),
```
neonphog, line 57: Turns out this was the bit causing my test flakes. The max_idle is now used as the timeout on attempting to connect to webrtc, which was enough for my local system but not CI.

jost-s (approved):
I'm not capable of reviewing every piece of the protocols. The general procedure of switching to webrtc first whenever possible and fall back to sbd makes a lot of sense to me.

ThetaSinner (approved):
Likewise not certain I'm getting all the details but I can follow the structure of what's changing and it looks good!

PR #131, opened 2025-03-25
Maint 03 25

ThetaSinner:
- The flake is pointing at `holochain/holochain`. If we want a flake, we can add one that points to `holochain/holonix`.
- A clippy fix.
- My IDE was highlighting that a bunch of dependencies weren't used. It doesn't always get it right but in this case `make` still works so I think it has a point.


Cargo.toml
```diff
@@ -14,50 +14,31 @@ resolver = "2"
...
 rand = "0.8.5"
-rand-utf8 = "0.0.1"
-rcgen = { version = "0.10.0", features = [ "pem" ] }
-ring = "0.16.20"
-rustls = { version = "0.20.0", features = [ "dangerous_configuration" ] }
-rustls-native-certs = "0.6.2"
-rustls-pemfile = "1.0.2"
-sbd-e2e-crypto-client = "0.0.9-alpha2"
-sbd-server = "0.0.9-alpha2"
+sbd-e2e-crypto-client = "0.0.8-alpha"
```
matthme, line 34: Is this downgrading to 0.0.8-alpha deliberately?

matthme: I'm gonna have to bump that soon anyway due to the [sbd PR](https://github.com/holochain/sbd/pull/42) though.

ThetaSinner: Not deliberate. This change has been sitting on my machine and got out-of-date. Fixed -> https://github.com/holochain/tx5/pull/131/commits/99ca1dfefcf43d121bca064d1cfcfc6bd9bb18e6

PR #133, opened 2025-03-27
bump to 0.3.1-beta

matthme:
Bumps all crates to 0.3.1-beta.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -54,7 +54,6 @@ jobs:
           key: avd-${{ matrix.android-api-level }}
 
       - name: Avd Snapshot
-        if: steps.avd-cache.outputs.cache-hit != 'true'
         env:
           ANDROID_API_LEVEL: ${{ matrix.android-api-level }}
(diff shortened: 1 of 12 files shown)


ThetaSinner (approved):

neonphog (approved):
heh, would be nice to switch to workspace version in this repo now that that's a thing. But doesn't need to be for this bump.

matthme:
Let's give that a try since we're at it.

.github/workflows/android.yml
```diff
@@ -54,7 +54,6 @@ jobs:
           key: avd-${{ matrix.android-api-level }}
 
       - name: Avd Snapshot
-        if: steps.avd-cache.outputs.cache-hit != 'true'
```
c12i, line 57: Curious why this is getting removed

matthme: Your curiosity is justified - I just removed this temporarily because the error in the android test said something about "cache needing deletion" and I was just trying whether that helps to resolve it. It does not so I just re-added it.

PR #137, opened 2025-04-24
feat!: Make configuration typed

ThetaSinner:
The JSON configuration matching the [MDN recommendation](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#configuration) really only worked with the Go pion backend. Datachannel uses a different configuration.

I'm proposing only exposing what is common. It does restrict the options that can be passed to the Go backend and requires an API change to add new options but I think what is gained by making the API clearer is worth that tradeoff.

### DIFF
--- a/crates/tx5-connection/src/config.rs
+++ b/crates/tx5-connection/src/config.rs
@@ -1,3 +1,4 @@
+use serde::{Deserialize, Serialize};
 use std::sync::Arc;
 
 /// The backend webrtc module to use.
@@ -34,4 +35,40 @@ pub struct HubConfig {
     /// Test falling back by failing webrtc setup.
     #[cfg(test)]
     pub test_fail_webrtc: bool,
+}
+
+/// Configuration for a group of ICE servers.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct IceServers {
+    /// The ICE server URLs to use for discovering external candidates.
+    pub urls: Vec<String>,
+}
+
+/// WebRTC config.
+///
+/// This configuration will be mapped the specific configuration used by
+/// the selected backend.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct WebRtcConfig {
+    /// A list of ICE servers configurations.
+    pub ice_servers: Vec<IceServers>,
+}
+
+#[cfg(feature = "backend-go-pion")]
+impl WebRtcConfig {
+    /// Convert this [`WebRtcConfig`] to a [`GoBuf`](tx5_go_pion::GoBuf).
+    pub fn to_go_buf(&self) -> std::io::Result<tx5_go_pion::GoBuf> {
+        serde_json::to_vec(self)
+            .map_err(|e| {
+                std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("failed to serialize WebRtcConfig: {}", e),
+                )
+            })?
+            .try_into()
+    }
 }
--- a/crates/tx5-connection/src/config.rs
+++ b/crates/tx5-connection/src/config.rs
@@ -37,6 +37,15 @@ pub struct HubConfig {
     pub test_fail_webrtc: bool,
 }
 
+/// The type of credential to use for ICE servers.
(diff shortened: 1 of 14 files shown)


jost-s (approved):
Looking good

crates/tx5-connection/src/config.rs
```diff
@@ -35,3 +36,39 @@ pub struct HubConfig {
...
+///
+/// This configuration will be mapped the specific configuration used by
+/// the selected backend.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct WebRtcConfig {
+    /// A list of ICE servers configurations.
+    pub ice_servers: Vec<IceServers>,
+}
```
neonphog, line 59:
```suggestion
pub struct WebRtcConfig {
    /// A list of ICE servers configurations.
    pub ice_servers: Vec<IceServers>,
    
    #[serde(default)]
    pub ice_transport_policy: TransportPolicy,
}

#[derive(Debug, Clone, Serialize, Deserialize, Default)]
#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
#[serde(rename_all = "camelCase")]
pub enum TransportPolicy {
    #[default]
    All,
    Relay,
}
```

I think we may also need the transport policy to be available for certain configurations.

neonphog (changes requested):
I think there are a couple more things that must be in there if we are locking this down. Otherwise looks good :+1:

crates/tx5-connection/src/webrtc/libdatachannel.rs
```diff
@@ -213,43 +213,27 @@ async fn task(
...
-        .port_range_begin(init_config.ephemeral_udp_port_min)
-        .port_range_end(init_config.ephemeral_udp_port_max);
+    let config = datachannel::RtcConfig::new::<String>(
+        &config
+            .ice_servers
+            .iter()
+            .flat_map(|s| s.urls.iter())
+            .cloned()
+            .collect::<Vec<_>>(),
+    )
```
neonphog, line 234: https://github.com/paullouisageneau/libdatachannel/blob/456c2151b7a1568b3805397a2d016e905c18a42e/src/configuration.cpp#L43-L87

Ugh... this is going to be frustrating to test. The rust bindings seem to only expose [the "url" constructor](https://github.com/paullouisageneau/libdatachannel/blob/456c2151b7a1568b3805397a2d016e905c18a42e/include/rtc/configuration.hpp#L23) for this config type.

So I guess we'll need to translate:

```json
iceServers: [{
  "urls": ["stun:one", "stun:two"],
  "credential": "abc",
  "username": "def",
}]
```

Into:

```json
iceServers: [
  "stun://abc:def@one",
  "stun://abc:def@two",
]
```

neonphog: This could be a follow-on at a later time.

crates/tx5-connection/src/config.rs
```diff
@@ -35,3 +36,39 @@ pub struct HubConfig {
...
 }
+
+/// Configuration for a group of ICE servers.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
+pub struct IceServers {
+    /// The ICE server URLs to use for discovering external candidates.
+    pub urls: Vec<String>,
+}
```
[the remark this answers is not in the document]
mattyg: Wouldn't including the credentials inline the url schema cover this?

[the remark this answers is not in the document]
ThetaSinner: It is possible, but not with the datachannel backend it seems

crates/tx5-connection/src/config.rs
ThetaSinner: Added

crates/tx5-connection/src/webrtc/libdatachannel.rs
ThetaSinner: Yeah I think given that we weren't doing anything with credentials here before it doesn't need to happen on this PR. I think this can probably wait until somebody asks for it though because at least for now, we don't need TURN.

crates/tx5-connection/src/config.rs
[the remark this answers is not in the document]
neonphog: I suppose technically, but the uri is not supposed to have credentials in it: https://www.rfc-editor.org/rfc/rfc7065#section-3.1

That is just a weird workaround libdatachannel implemented.

neonphog (approved):

PR #142, opened 2025-05-08
feat: Get connected URLs from the endpoint

ThetaSinner:
Supporting change for closing connections in K2. The only other option was to get peer ids through the stats method. But then I'd have to carefully reconstruct URLs and potentially make a mess with multiple signal servers being in use. Better to just expose this I think.

### DIFF
--- a/crates/tx5-go-pion-sys/src/lib.rs
+++ b/crates/tx5-go-pion-sys/src/lib.rs
@@ -315,7 +315,6 @@ impl Api {
                     let msg =
                         std::slice::from_raw_parts(slot_b as *const u8, slot_c);
                     let msg = String::from_utf8_lossy(msg);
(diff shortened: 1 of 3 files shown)


zippy (approved):
very straightforward, looks good.

Does it make sense to have test that shows that it works when an ep is connected to more than one url?

neonphog (approved):
:+1:

ThetaSinner:
> Does it make sense to have test that shows that it works when an ep is connected to more than one url?

Could do. If there was any filtering or logic involved then I would say yes. For me, the main reason for a test for a method that is simple like this one is to make somebody realize that somebody cared enough to add a test in the future, and check what the function was used for before deciding it was unused and deleting it :)

PR #144, opened 2025-05-09
Ensure CI passes for release-0.1.x-beta branch

neonphog:


.github/workflows/static.yml
```diff
@@ -20,7 +23,6 @@ jobs:
         ]
         toolchain: [
           stable,
-          1.81.0
```
matthme, line 23: holochain's main-0.4 branch still has 1.81.0 in its [rust-toolchain.toml](https://github.com/holochain/holochain/blob/71a763fd12bba5dc7c12b3d41cc4a6430f1c2004/rust-toolchain.toml#L2) that's why I think it was here. Is it safe to ignore that?

neonphog: There was some incompatibility with the rdme doc generation tool and that static check... I'm inclined to assume it won't be a problem, and if it is, we can re-instate this and fix that other issue.

matthme (approved):

PR #150, opened 2025-05-21
add doc comments for tx5-connection

neonphog:


matthme (changes requested):
Is [the doc comment in lib.rs](https://github.com/holochain/tx5/blob/5ffc256865059c9eb476b745450a5985793ada02/crates/tx5-connection/src/lib.rs#L6-L25) still up to date? It's saying that go pion is the default implementation but I thought we switched to datachannel now, right?

neonphog:
@matthme 

> Is [the doc comment in lib.rs](https://github.com/holochain/tx5/blob/5ffc256865059c9eb476b745450a5985793ada02/crates/tx5-connection/src/lib.rs#L6-L25) still up to date? It's saying that go pion is the default implementation but I thought we switched to datachannel now, right?

yep, good call, missed that one: https://github.com/holochain/tx5/pull/150/commits/4c57e280e84863446b31ab45fc436ef21573dda8

jost-s (approved):

PR #152, opened 2025-05-28
2 passing and 2 ignored failing event tests

neonphog:


crates/tx5/Cargo.toml
```diff
@@ -28,6 +28,9 @@ backend-webrtc-rs = [ "tx5-connection/backend-webrtc-rs" ]
 # make a schema available for configuration types
 schema = [ "tx5-connection/schema" ]
 
+# enable some testing utility apis
+test_utils = [ "tx5-connection/test_utils" ]
```
ThetaSinner, line 32: There's no naming standard for these but we are more often using kebab case. In Holochain it's a mixture and the test utils are mostly snake case. I think as this case is already using kebab case, I'd prefer kebab case for this feature.

Just checked K2, we managed to keep it consistently kebab case in there too.

crates/tx5/tests/tests/events.rs
```diff
@@ -0,0 +1,199 @@
+use std::sync::Arc;
+
+async fn sbd() -> sbd_server::SbdServer {
+    let config = sbd_server::Config {
+        bind: vec!["127.0.0.1:0".to_string(), "[::1]:0".to_string()],
+        limit_clients: 100,
+        disable_rate_limiting: true,
```
ThetaSinner, line 7: Should we have done this on the bootstrap2 server? I thought the other day when we were talking about 0.4 vs 0.5 that we hadn't done that with the new bootstrap servers but it slipped my mind again.

ThetaSinner (commented):
Sweet, this is super helpful, thank you. Minor queries from me, the tests are great

jost-s (dismissed):
Sweet!

matthme (dismissed):
Great! Small suggestion, provided that I associated the tests that are currently ignored to the correct github issue.

matthme (approved):

PR #158, opened 2025-07-16
chore: Add a new signal integration test and refactor some nearly-duplicated code

ThetaSinner:
- Add a new test that can be tuned to make signal demonstrate rate limiting. By enabling, disabling or lowering rate limiting you can see how the runtime of the test is affected
- Add some extra tracing
- Refactored a bunch of nearly-duplicated code for starting an SBD server and creating an endpoint.

### DIFF
--- a/crates/tx5/tests/tests/mod.rs
+++ b/crates/tx5/tests/tests/mod.rs
@@ -1,3 +1,58 @@
+use std::sync::Arc;
+use tx5::{Endpoint, EndpointRecv, PeerUrl};
+
 mod events;
 mod multi_sig;
+mod relay_over_sig;
 mod stress;
+
+async fn sbd() -> sbd_server::SbdServer {
+    let config = sbd_server::Config {
+        bind: vec!["127.0.0.1:0".to_string(), "[::1]:0".to_string()],
+        limit_clients: 100,
+        disable_rate_limiting: true,
+        ..Default::default()
+    };
+    sbd_with_config(config).await
+}
+
+async fn sbd_with_config(
+    mut config: sbd_server::Config,
+) -> sbd_server::SbdServer {
+    // Always configure the bind addresses for tests
+    config.bind = vec!["127.0.0.1:0".to_string(), "[::1]:0".to_string()];
+    sbd_server::SbdServer::new(Arc::new(config)).await.unwrap()
+}
+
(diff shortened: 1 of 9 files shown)


crates/tx5/tests/tests/relay_over_sig.rs
```diff
@@ -0,0 +1,102 @@
...
+    let msg = receive_next_message_from(&mut r2, p1.clone()).await;
+    assert_eq!(large_msg, msg);
+
+    e2.send(p1.clone(), large_msg.clone()).await.unwrap();
+    let msg = receive_next_message_from(&mut r1, p2.clone()).await;
+    assert_eq!(large_msg, msg);
+
+    // Send a batch of large messages
+
+    let end = Instant::now() + std::time::Duration::from_secs(60);
```
cdunster, line 64: Nitpick: This is only used for the tokio `timeout_at`, so it would be better to move its definition closer to its use.

jost-s (approved):
Yeah

crates/tx5/tests/tests/relay_over_sig.rs
cdunster: Ah, yes, clear. That does make sense, but I would also expect sending to be a negligible amount of time.

crates/tx5/tests/tests/mod.rs
```diff
@@ -1,3 +1,60 @@
...
+mod relay_over_sig;
 mod stress;
+
+async fn sbd() -> sbd_server::SbdServer {
+    let config = sbd_server::Config {
+        bind: vec!["127.0.0.1:0".to_string(), "[::1]:0".to_string()],
+        limit_clients: 100,
+        disable_rate_limiting: true,
+        // limit_ip_kbps: 100_000,
+        // limit_ip_byte_burst: 26000000,
```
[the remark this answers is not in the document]
ThetaSinner: Removed! This was from my debuggng and should have been removed yes

crates/tx5/tests/tests/relay_over_sig.rs
matthme: It could be useful to add some of this context as an inline comment. I think I wouldn't have understood that easily without reading above discussion.

matthme (approved):

PR #159, opened 2025-07-16
chore: Expose WebRTC fail

ThetaSinner:


zippy (approved):

cdunster (commented):
Why not just enable the `test-utils` feature in K2 and Holochain when we want it? Is the idea that we always want this enabled in production as well? If so, then I think the field should be renamed to remove the `test_` prefix.

jost-s (approved):
Great!

cdunster (approved):

Issue #161, opened 2025-07-22
Ensure connections close correctly on WebRTC disconnection


ThetaSinner:
See point 4 of the parent issue.

PR #162, opened 2025-07-22
chore: Close connections when an established WebRTC connection fails

ThetaSinner:
This allows the WebRTC connection to be re-established and if that doesn't work then the connection establishment can fall back to the relay. Without this change, peers that could otherwise be using WebRTC can end up running over the relay.

Closes https://github.com/holochain/tx5/issues/161


crates/tx5-connection/src/conn.rs
```diff
@@ -367,6 +367,7 @@ async fn webrtc_task(
             break;
         }
     }
+    netaudit!(DEBUG, a = "webrtc task closed, sending WebrtcClosed",);
```
ThetaSinner, line 370: When connections do close, we still want to know why and currently don't have any information to help trace where the disconnection starts. I've added a bit of extra tracing where I think we're missing information.

crates/tx5-connection/src/conn.rs
```diff
@@ -623,7 +624,7 @@ async fn con_task_attempt_webrtc(
                     pub_key = ?task_core.pub_key,
                     a = "webrtc fallback: webrtc processing task closed",
                 );
-                return Fallback(task_core);
+                break;
```
ThetaSinner, line 627: The fix. We were deliberately choosing to fall back to the relay when WebRTC fails after being established.

This is what I mean when I complement neonphog's code. You come in looking for an issue and 99% of the code is sensible and you're just looking for a detail to tweak. In this case, the code was expressing exactly what we were seeing happen and not what we want to have happen. Easy to fix.

crates/tx5/tests/tests/mod.rs
```diff
@@ -46,6 +47,30 @@ async fn ep_with_config(
     (peer_url, ep, recv)
 }
 
+async fn receive_next_message_from(
```
ThetaSinner, line 50: Extracted common code from an existing test.

crates/tx5/tests/tests/reconnect.rs
```diff
@@ -0,0 +1,62 @@
+use crate::tests::{
+    enable_tracing, ep, ep_with_config, receive_next_message_from, sbd,
+    sbd_with_config,
+};
+use tokio::time::Instant;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn reconnect_on_webrtc_fail() {
```
ThetaSinner, line 8: - Demonstrate that we can get a WebRTC connection by sending messages. 
- Assert that we do in fact have one WebRTC connection each
- Drop one side to force a shutdown. No expectation that this sends a notification to the other side to tell it we're shutting down. That only happens if we explicitly close the connection before dropping it.
- Assert that the other peer goes to no connections after trying to send a message to the unavailable peer.
- Bring the dropped peer "back" with a new peer address as though they'd restarted their program - and check that we re-open a connection to them.

It's not just important that we close connections and remove them from state, but that enough state is gone that we can actually re-establish a connection to them. The test isn't perfect in that regard and it would be nice to re-connect to the exact same peer address but I think that'd take a bunch more fiddling to make the test do that and this already demonstrates that the fix works.

If you run the test with the fix above reverted then it fails.

matthme (dismissed):
Looks great!

crates/tx5-connection/src/conn.rs
matthme: I think this `break;` is in effect equivalent to `return Abort` as is used in the other match arms. So we could potentially use `return Abort` here as well for a bit more consistency?

ThetaSinner: Agree that consistency is good but went the other way and changed all the aborts to break statements. It makes sense to return the fallback because you really want to redirect the control flow but for the termination I think it's a little clearer to let the function abort by finishing.

crates/tx5-connection/src/conn.rs
```diff
@@ -623,7 +624,7 @@ async fn con_task_attempt_webrtc(
                     pub_key = ?task_core.pub_key,
                     a = "webrtc fallback: webrtc processing task closed",
```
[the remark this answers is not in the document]
ThetaSinner: It does, updated

jost-s (approved):
Super

PR #164, opened 2025-08-06
Debug unusable connection in state

ThetaSinner:


crates/tx5/tests/tests/investigate.rs
```diff
@@ -0,0 +1,95 @@
...
+    let (p2, e2, mut r2) = ep(&sig).await;
+    println!("[2] Listening on: {:?}", p2);
+
+    e1.send(p2.clone(), b"hello".to_vec()).await.ok();
+    let got = receive_next_message_from(&mut r2, p1).await;
+    tracing::info!("Received message: {:?}", String::from_utf8_lossy(&got));
+
+    tracing::info!("Dropping all connections...");
+
+    sig.drop_all_connections().await;
```
ThetaSinner, line 24: Now force connections to close to the signal server remotely

crates/tx5/tests/tests/investigate.rs
```diff
@@ -0,0 +1,95 @@
...
+    tracing::info!("All connections dropped, waiting for 5 seconds...");
+
+    // Close the connection to p2, so that we'll need to reconnect over the missing signal server.
+    e1.close(&p2);
```
ThetaSinner, line 28: Close the direct connection to peer 2

crates/tx5/src/backend/be_tx5_connection.rs
```diff
@@ -50,7 +52,22 @@ impl BackWaitCon for GoWaitCon {
...
-            con.ready().await;
+            tracing::info!(
+                "Waiting for connection to be ready for pub_key: {:?}",
+                pub_key
+            );
+            // This connection will only ready on a code path that successfully establishes either
+            // a WebRTC or a relayed connection. However, there are many exits from the code paths
+            // that attempt to set those up. In those cases, we own a semaphore permit that will
+            // never be released and we risk deadlocking here without this timeout.
+            tokio::time::timeout(timeout, con.ready())
```
ThetaSinner, line 63: This ready check requires a semaphore to be dropped. This only happens if we successfully establish a network connection or fall back to the relay.

If something goes wrong here, we can end up waiting indefinitely for an event that won't happen. That is what leaves the connection stuck in a state where no messages sent.

If something does go wrong and we know about it, it'd be nice to notify this ready check immediately with an error... I will have a think about that on the coming PR that contains just the fix and not my debugging.

PR #165, opened 2025-08-06
fix: Connections should close if the signal server is discovered to be unavailable during connection setup or the connection setup otherwise fails

ThetaSinner:


crates/tx5/src/config.rs
```diff
@@ -57,6 +57,16 @@ pub struct Config {
 
     /// Test falling back to the signal relay by failing the WebRTC setup.
     pub danger_force_signal_relay: bool,
+
+    /// Deny using the signal server as a relay if direct connections fail.
+    ///
+    /// This is useful if you want to ensure that the signal server is only used for connection
+    /// establishment and not as a relay.
+    ///
+    /// Enabling the setting is dangerous, though as it will mean that if a peer fails to make a
```
matthme, line 66:
```suggestion
    /// Enabling the setting is dangerous though, as it will mean that if a peer fails to make a
```

matthme (approved):
Looks good as far as I can tell :+1:

crates/tx5-connection/src/hub.rs
```diff
@@ -56,7 +56,7 @@ async fn hub_map_assert(
 
     client.assert(&pub_key).await?;
 
-    // we're connected to the peer, create a connection
+    // we're not connected to the peer, create a connection
```
jost-s, line 59: important distinction

crates/tx5/src/backend/be_tx5_connection.rs
```diff
@@ -50,7 +51,17 @@ impl BackWaitCon for GoWaitCon {
                 _ => return Err(std::io::Error::other("already awaited")),
             };
 
-            con.ready().await;
+            // This connection will only ready on a code path that successfully establishes either
+            // a WebRTC or a relayed connection. However, there are many exits from the code paths
+            // that attempt to set those up. In those cases, we own a semaphore permit that will
+            // never be released and we risk deadlocking here without this timeout.
+            tokio::time::timeout(timeout, con.ready())
```
jost-s, line 58: 💯

crates/tx5/src/ep.rs
```diff
@@ -189,6 +189,11 @@ impl std::fmt::Debug for Endpoint {
 impl Endpoint {
     /// Construct a new tx5 endpoint.
     pub fn new(config: Arc<Config>) -> (Self, EndpointRecv) {
+        if config.danger_force_signal_relay && config.danger_deny_signal_relay {
+            #[cfg(not(feature = "test-utils"))]
+            tracing::warn!("Both `danger_force_signal_relay` and `danger_deny_signal_relay` are set to true. This is likely a misconfiguration, as these options are mutually exclusive.");
+        }
```
jost-s, line 195:
```suggestion
        #[cfg(not(feature = "test-utils"))]
        if config.danger_force_signal_relay && config.danger_deny_signal_relay {
            tracing::warn!("Both `danger_force_signal_relay` and `danger_deny_signal_relay` are set to true. This is likely a misconfiguration, as these options are mutually exclusive.");
        }
```

jost-s (approved):
Fantastic find!

crates/tx5/src/config.rs
ThetaSinner: Updated

PR #173, opened 2025-08-13
perf: Drop outbound messages when the connection closes, without waiting for a timeout

ThetaSinner:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of peer connections and message delivery.
  * Sends now return a clear error when a connection can’t be established.
  * Broadcast skips unreachable peers without blocking other sends.
  * Reduced risk of stalls during connection setup; dropped messages are logged for diagnostics.

* **Refactor**
  * Streamlined connection readiness handling for lower overhead and faster responsiveness.

### DIFF
--- a/crates/tx5/src/ep.rs
+++ b/crates/tx5/src/ep.rs
@@ -134,10 +134,6 @@ impl EpInner {
 
     /// Get an existing peer connection or create a new outgoing one.
     pub fn connect_peer(&mut self, peer_url: PeerUrl) -> Arc<Peer> {
-        if let Some(peer) = self.peer_map.get(&peer_url) {
-            return peer.clone();
-        }
-
         self.peer_map
             .entry(peer_url.clone())
             .or_insert_with(|| {
@@ -238,8 +234,8 @@ impl Endpoint {
 
     /// Send data to a remote on this tx5 endpoint.
     ///
-    /// The future returned from this method will resolve when
-    /// the data is handed off to our networking backend.
+    /// The future returned from this method will resolve when the data is handed off to the
+    /// networking backend or the connection attempt to the peer has failed.
     pub async fn send(&self, peer_url: PeerUrl, data: Vec<u8>) -> Result<()> {
         let listening_addresses = self.get_listening_addresses();
         if listening_addresses.contains(&peer_url) {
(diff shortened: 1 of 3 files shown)


crates/tx5/src/ep.rs
```diff
@@ -272,8 +272,9 @@ impl Endpoint {
                 let data = data.to_vec();
                 tokio::task::spawn(async move {
                     let _ = tokio::time::timeout(timeout, async {
-                        peer.ready().await;
-                        let _ = peer.send(data).await;
+                        if let Some(conn) = peer.wait_for_ready().await {
+                            let _ = conn.send(data).await;
+                        }
```
ThetaSinner, line 277: We don't log individual errors here on broadcast. Again, debatable but also potentially very noisy on a larger network.

crates/tx5/src/peer.rs
```diff
@@ -101,41 +97,26 @@ impl Peer {
...
-    pub async fn send(&self, msg: Vec<u8>) -> Result<()> {
-        let conn = match &*self.ready.lock().unwrap() {
-            MaybeReady::Ready(c) => c.clone(),
-            _ => return Err(Error::other("not ready")),
-        };
-        conn.send(msg).await
+    /// This future resolves when the connection is ready to use or has failed to connect.
+    ///
+    /// If the connection is not usable, it will return `None`.
+    pub async fn wait_for_ready(&self) -> Option<DynBackCon> {
```
ThetaSinner, line 118: Note the change here. I'm slightly breaking encapsulation by handing out the connection but it is only handed out within the module. It's given back to the endpoint to use.

By having the "ready" operation return the connection when the connection is available we can just have one function and let the caller do their own send.

crates/tx5/src/peer/maybe_ready.rs
```diff
@@ -0,0 +1,207 @@
...
+                w.close();
+            }
+            MaybeReadyState::Failed => {
+                tracing::warn!(
+                    "Cannot set state to ready because state is already failed"
+                );
+                return;
+            }
+            MaybeReadyState::Ready(_) => {
+                tracing::error!(
```
matthme, line 45: What's the reason for using `error!` here and `warn!` above in the `::Failed` case?

crates/tx5/src/ep.rs
```diff
@@ -134,10 +134,6 @@ impl EpInner {
 
     /// Get an existing peer connection or create a new outgoing one.
     pub fn connect_peer(&mut self, peer_url: PeerUrl) -> Arc<Peer> {
-        if let Some(peer) = self.peer_map.get(&peer_url) {
-            return peer.clone();
-        }
-
```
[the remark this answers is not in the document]
cdunster: I guess that it avoids the unnecessary `peer_url.clone()` if there is an entry present, but I would hope that the compiler is smart enough to make this a move in this case, and if not, then it's not a big deal.

cdunster (approved):

matthme (commented):
The logic looks good to me, I just have a request to update docs in one place and a question about tracing levels.

PR #182, opened 2025-08-22
test: verify that a flooded connection closes the channel correctly

### LINKED ISSUE #176: Verify that a flooded connection closes correctly
When tx5 receives more messages from the app or from the network than it can handle, it kills the channel that relays messages between the network and the application. Write tests to verify that the channel closing results in the connection shutting down.

See the warning added to notify us if this happens: https://github.com/holochain/tx5/pull/175/files#diff-bb2fb84dc37b8786cdcab5aae7d5da7438d441ae3ffaceb7bdc6f0a568cf6668R98

cdunster:
Closes #176


crates/tx5-connection/Cargo.toml
```diff
@@ -47,3 +47,5 @@ schemars = { workspace = true, optional = true }
 rand = { workspace = true }
 sbd-server = { workspace = true }
 tracing-subscriber = { workspace = true }
+tracing-test = "0.2.5"
```
cdunster, line 50: If we don't want to use the log asserts, then there is the more widely used [test-log](https://crates.io/crates/test-log) crate.

crates/tx5-connection/src/test.rs
```diff
@@ -410,3 +390,104 @@ async fn framed_con_drop_disconnects() {
...
+    let res = tx.send(b"hello".to_vec()).await;
+    assert!(
+        matches!(res, Err(ref e) if e.kind() == ErrorKind::BrokenPipe),
+        "Expected BrokenPipe error, got: {res:?}"
+    );
+
+    // A warning is logged that the message failed to send and the channel will be closed.
+    assert!(logs_contain(
+        "WARN channel_is_closed_on_send_when_flooded: tx5_connection: Failed to send message, closing channel"
+    ));
```
cdunster, line 412: It might be unnecessary to assert this, but a `BrokenPipe` error can be returned for any reason, so this makes sure that we have the correct code branch.

ThetaSinner (changes requested):
- As unit tests of the library code for the DropChannel type I think these are fine. I haven't looked too closely because of the `tracing-test` use. I think the tests demonstrate a good amount of observable behavior without the library. I've been burned too many times by libraries with [so little track record for maintenance](https://github.com/dbrgn/tracing-test), as well as not being keen on the idea of asserting logs when there are observable effects already tested.
- Although the unit tests cover the drop channel types, this isn't what the issue was about. The issue was about setting up full tx5 connections that have this type embedded in them and then causing the channel to flood. At that point, we want to make sure that the receive or send side being flooded will cause the whole connection to shut down and allow a new connection to replace it.

PR #184, opened 2025-08-26
fix: incompatible WebRtcConfig for go-pion

ThetaSinner:
It wasn't actually possible to configure the go-pion backend to make p2p connections except on a local network. This is only affecting 0.6 currently.

## Summary by CodeRabbit

* **Improvements**
  * Serialized configuration now omits empty ICE server credential fields, producing cleaner outputs.
  * Transport policy values in serialized configuration now use camelCase for improved consistency.

* **Potential Impact**
  * Tools or scripts that relied on always-present (but empty) credential fields or on previous casing for policy values may require updates.

### DIFF
--- a/crates/tx5-connection/src/config.rs
+++ b/crates/tx5-connection/src/config.rs
@@ -57,21 +57,22 @@ pub struct IceServers {
     pub urls: Vec<String>,
 
     /// The username to use for authentication.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub username: Option<String>,
 
     /// The credential to use for authentication.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub credential: Option<String>,
 
     /// The credential type to use for authentication.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub credential_type: Option<CredentialType>,
 }
 
 /// ICE transport policy.
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


crates/tx5-connection/src/config.rs
```diff
@@ -57,21 +57,22 @@ pub struct IceServers {
...
     /// The credential type to use for authentication.
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub credential_type: Option<CredentialType>,
 }
 
 /// ICE transport policy.
 #[derive(Debug, Clone, Default, Serialize, Deserialize)]
 #[cfg_attr(feature = "schema", derive(schemars::JsonSchema))]
+#[serde(rename_all = "camelCase")]
```
ThetaSinner, line 75: It's unclear whether datachannel actually uses this value, but for go-pion the case needed fixing https://github.com/pion/webrtc/blob/master/icetransportpolicy.go#L27-L30

jost-s (approved):
That ICE configuration isn't exposed downstream, is it?

matthme (approved):

crates/tx5-connection/src/config.rs
ThetaSinner: We would not necessarily, because it doesn't need a valid ICE configuration. Everything works fine on a local network but falls apart when you need to know your public IP

ThetaSinner:
> Then it's a breaking change, right? But doesn't matter on 0.6

It technically is a breaking change but it's only breaking for people who were passing configuration that wasn't valid for go-pion anyway... so I think it's acceptable. We'll see if cargo-semver-checks agrees

PR #186, opened 2025-08-27
fix: Time out the WebRTC connection before timing out the connection attempt

ThetaSinner:
Tx5 sets up several timeouts on the connection path:
- The message send https://github.com/holochain/tx5/blob/main/crates/tx5/src/ep.rs#L246
- The connection attempt https://github.com/holochain/tx5/blob/main/crates/tx5/src/peer.rs#L230
- The WebRTC connection attempt https://github.com/holochain/tx5/blob/e041113abe323d8bc3f19af758f8f2cf91b98540/crates/tx5-connection/src/conn.rs#L400

The first two being the same, means that the message send gives up before the connection attempt gives up. This is only by a small amount in most cases and isn't especially serious i think.

The second pair is really bad. The connection attempt will always timeout before it gives up on connecting WebRTC. As a consequence, we can never use the relay. 

This bug already existed but I've effectively revealed it with the other fixes.

The change can't be backported like this, I'll backport a simpler version of the fix.

## Summary by CodeRabbit

- **New Features**
  - Configurable WebRTC connection timeout (default 45s) and builder setter.
  - Options to force or deny signal relay.
  - Config is cloneable with a builder-style API and convenience setters.

- **Refactor**
  - Examples, benchmarks, and tests migrated to the new builder-based configuration.

- **Tests**
  - Relay test renamed and new fallback-after-WebRTC-timeout scenario added.
  - Reconnect and fail-WeRTC tests updated for stricter cleanup and behavior verification.

### DIFF
--- a/crates/tx5-connection/src/config.rs
+++ b/crates/tx5-connection/src/config.rs
@@ -32,6 +32,14 @@ pub struct HubConfig {
     /// The signal config to use.
     pub signal_config: Arc<tx5_signal::SignalConfig>,
 
+    /// The timeout for establishing a WebRTC connection to a peer.
+    ///
+    /// After this time has elapsed, if a WebRTC connection has not been established, the connection
+    /// will fall back to using the signal server as a relay. Any timeout applied to the
+    /// [`Conn::ready`] or [`Conn::send`] futures should allow enough time for this timeout to
+    /// elapse and for the message to be sent over the relay.
+    pub webrtc_connect_timeout: std::time::Duration,
+
     /// Test falling back to the signal relay by failing the WebRTC setup.
     pub danger_force_signal_relay: bool,
 
--- a/crates/tx5-connection/src/conn.rs
+++ b/crates/tx5-connection/src/conn.rs
@@ -394,7 +394,7 @@ async fn con_task_attempt_webrtc(
 ) -> AttemptWebrtcResult {
     use AttemptWebrtcResult::*;
 
-    let timeout_dur = task_core.config.signal_config.max_idle;
+    let timeout_dur = task_core.config.webrtc_connect_timeout;
     let timeout_cmd_send = task_core.cmd_send.clone();
     tokio::task::spawn(async move {
         tokio::time::sleep(timeout_dur).await;
--- a/crates/tx5/src/backend/be_tx5_connection.rs
+++ b/crates/tx5/src/backend/be_tx5_connection.rs
@@ -140,7 +140,6 @@ pub async fn connect(
             ..Default::default()
         },
         listener,
-        //max_connections: config.connection_count_max as usize,
         max_idle: config.timeout,
         ..Default::default()
     };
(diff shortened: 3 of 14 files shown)


crates/tx5-connection/src/config.rs
```diff
@@ -32,6 +32,9 @@ pub struct HubConfig {
     /// The signal config to use.
     pub signal_config: Arc<tx5_signal::SignalConfig>,
 
+    /// The timeout for establishing a WebRTC connection to a peer.
+    pub webrtc_connect_timeout: std::time::Duration,
```
ThetaSinner, line 36: We can't just have a single timeout config any more. Adding a new configuration lets the WebRTC connection timeout be set separately to the overall timeout.

crates/tx5-connection/src/conn.rs
```diff
@@ -394,7 +394,7 @@ async fn con_task_attempt_webrtc(
 ) -> AttemptWebrtcResult {
     use AttemptWebrtcResult::*;
 
-    let timeout_dur = task_core.config.signal_config.max_idle;
```
ThetaSinner, line 397: This `max_idle` is set to the same value as `Config::timeout` by the calling code.

crates/tx5/src/config.rs
```diff
@@ -3,54 +3,99 @@ use tx5_connection::WebRtcConfig;
 use tx5_core::deps::serde_json;
 
 /// Tx5 endpoint configuration.
+#[derive(Clone)]
+#[non_exhaustive]
```
ThetaSinner, line 7: This is a breaking change now, and this is the part I can't backport. I think it's worth doing now because it means we can add new config parameters without it being a breaking change. I wish we'd done it sooner so adding `webrtc_connect_timeout` wouldn't be a breaking change.

crates/tx5-connection/src/config.rs
jost-s: I'm wondering whether it's better to not expose this setting, and instead subtract some fixed value from the connection timeout, with a minimum, to prevent this situation.
If you don't think it is, then I'd document here that it prevents falling back to relay when exceeding the connection timeout.

matthme: I tend to agree with @jost-s . Otherwise, people may set the `webrtc_connect_timeout` larger than the overall timeout without being aware of the consequences.

matthme: I just saw the documentation in the `Config` struct below and it's quite elaborate so I'm probably fine leaving the variable in the config. But it might still be documented better on the `HubConfig` as well here.

matthme (dismissed):
Looks good :+1: Just a small comment on the `HubConfig`.

crates/tx5-connection/src/config.rs
ThetaSinner: The overall timeout only exists in `tx5`, not in the `tx5-connection` crate. I have added a general comment to this configuration setting to ask that the `ready` and `send` methods on the `Conn` struct are given enough time for WebRTC to time out and for the connection to fall back to the relay and send the message.

crates/tx5/src/backend/be_tx5_connection.rs
```diff
@@ -162,6 +161,7 @@ pub async fn connect(
     let hub_config = Arc::new(tx5_connection::HubConfig {
         backend_module,
         signal_config: Arc::new(sig_config),
+        webrtc_connect_timeout: config.webrtc_connect_timeout,
         danger_force_signal_relay: config.danger_force_signal_relay,
         danger_deny_signal_relay: config.danger_deny_signal_relay,
     });
```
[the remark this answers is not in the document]
ThetaSinner: This is a good suggestion. I added a warning to be logged if the `webrtc_connect_timeout` is set to 0 and modified the `headroom` variable name - but the change is largely just accepted.

jost-s (approved):

PR #194, opened 2025-10-31
fix: drop connection on pion event that PeerConnectionState has changed to disconnected, closed, or failed

mattyg:
With the pion backend, connections are now closed upon receiving an event that the `RTCPeerConnection` state has changed to `disconnect`, `closed` or `failed`.

The spec outlines these states: https://w3c.github.io/webrtc-pc/#rtcpeerconnectionstate-enum. I believe this is the right approach for handling them. This is how it is already implemented in the libdatachannel backend.

## Summary by CodeRabbit

* **Bug Fixes**
  * Terminal WebRTC states and data-channel closures now emit explicit close signals (not errors), stop tasks reliably, and ignore spurious closed events during negotiation.

* **New Features**
  * Added a distinct "Closed" WebRTC event plus explicit Close and Error command signals for clearer connection-state visibility.

* **Refactor**
  * Peer lifecycle moved into an inner task with unified event/command forwarding to simplify termination handling.

* **Public API**
  * Command and event enums updated to distinguish closed vs error outcomes.

* **Tests**
  * Added tests validating terminal peer-state transitions and closure handling.

### DIFF
--- a/crates/tx5-connection/src/webrtc/go_pion.rs
+++ b/crates/tx5-connection/src/webrtc/go_pion.rs
@@ -16,6 +16,8 @@ enum Cmd {
     RecvMessage(Vec<u8>),
     DataChanOpen,
     BufferedAmountLow,
+    Close,
+    Error(std::io::Error),
 }
 
 pub struct Webrtc {
@@ -107,14 +109,36 @@ async fn task(
     is_polite: bool,
     config: WebRtcConfig,
     send_buffer: usize,
+    evt_send: CloseSend<WebrtcEvt>,
+    cmd_send: CloseSend<Cmd>,
+    cmd_recv: CloseRecv<Cmd>,
+) -> Result<()> {
+    let config = config.to_go_buf()?;
+    let (peer, peer_evt) = tx5_go_pion::PeerConnection::new(config).await?;
+    task_inner(
+        is_polite,
+        send_buffer,
+        evt_send,
+        cmd_send,
+        cmd_recv,
+        peer,
+        peer_evt,
+    )
+    .await
+}
+
+async fn task_inner(
+    is_polite: bool,
+    send_buffer: usize,
     mut evt_send: CloseSend<WebrtcEvt>,
     cmd_send: CloseSend<Cmd>,
     mut cmd_recv: CloseRecv<Cmd>,
+    peer: tx5_go_pion::PeerConnection,
+    mut peer_evt: tokio::sync::mpsc::UnboundedReceiver<
+        tx5_go_pion::PeerConnectionEvent,
+    >,
 ) -> Result<()> {
     evt_send.set_close_on_drop(true);
-
-    let config = config.to_go_buf()?;
-    let (peer, mut peer_evt) = tx5_go_pion::PeerConnection::new(config).await?;
 
     let mut cmd_send2 = cmd_send.clone();
     let _peer_task: AbortTask<Result<()>> = AbortTask(tokio::task::spawn(
@@ -124,8 +148,33 @@ async fn task(
             use tx5_go_pion::PeerConnectionEvent as Evt;
             while let Some(evt) = peer_evt.recv().await {
                 match evt {
-                    Evt::Error(_) => break,
-                    Evt::State(_) => (),
+                    Evt::Error(err) => {
+                        cmd_send2.send_or_close(Cmd::Error(
+                            std::io::Error::other(format!(
+                                "Pion error event: {err}"
+                            )),
+                        ))?;
+                        break;
+                    }
+                    Evt::State(state) => match state {
+                        tx5_go_pion::PeerConnectionState::Disconnected => {
+                            cmd_send2.send_or_close(Cmd::Close)?;
+                            break;
+                        }
+                        tx5_go_pion::PeerConnectionState::Closed => {
+                            cmd_send2.send_or_close(Cmd::Close)?;
+                            break;
+                        }
+                        tx5_go_pion::PeerConnectionState::Failed => {
+                            cmd_send2.send_or_close(Cmd::Error(
+                                std::io::Error::other(
+                                    "PeerConnectionState changed to Failed",
+                                ),
+                            ))?;
+                            break;
+                        }
+                        _ => (),
+                    },
                     Evt::ICECandidate(mut ice) => {
                         cmd_send2
                             .send_or_close(Cmd::GeneratedIce(ice.to_vec()?))?;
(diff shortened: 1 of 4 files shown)


crates/tx5-connection/src/webrtc/go_pion.rs
```diff
@@ -107,15 +108,37 @@ async fn task(
...
+        evt_send,
+        cmd_send,
+        cmd_recv,
+        peer,
+        peer_evt,
+    )
+    .await
+}
+
+async fn task_inner(
```
mattyg: Move logic into inner function so that we can access the `PeerConnection` in test

crates/tx5-connection/src/webrtc/go_pion.rs
```diff
@@ -125,6 +148,29 @@ async fn task(
...
+                            ),
+                        ))?;
+                    }
+                    Evt::State(tx5_go_pion::PeerConnectionState::Failed) => {
+                        cmd_send2.send_or_close(Cmd::Error(
+                            std::io::Error::other(
+                                "PeerConnectionState changed to Failed",
+                            ),
+                        ))?;
+                    }
```
mattyg: The actual change.

crates/tx5-connection/src/webrtc/go_pion.rs
```diff
@@ -269,3 +316,94 @@ fn spawn_data_chan(
...
+        });
+
+        // Send connected state
+        peer_evt_send
+            .send(tx5_go_pion::PeerConnectionEvent::State(
+                tx5_go_pion::PeerConnectionState::Connected,
+            ))
+            .unwrap();
+
+        assert!(!task_handle.is_finished());
```
matthme: Shouldn't we in principle wait for a while here before checking, since sending over the channel may take a while?

crates/tx5-connection/src/webrtc/go_pion.rs
```diff
@@ -125,6 +148,29 @@ async fn task(
             while let Some(evt) = peer_evt.recv().await {
                 match evt {
                     Evt::Error(_) => break,
```
matthme: I'm noticing that this `Error` event only breaks the loop while the states `Disconnected`, `Closed` and `Failed` send an error through the channel. I wonder whether they should be treated equally? But I have to admit that I don't understand the surrounding overall logic yet.

crates/tx5-connection/src/webrtc/go_pion.rs
```diff
@@ -124,8 +147,39 @@ async fn task(
...
+                                    ),
+                                ))?;
+                            },
+                            tx5_go_pion::PeerConnectionState::Closed => {
+                                cmd_send2.send_or_close(Cmd::Error(
+                                    std::io::Error::other(
+                                        "PeerConnectionState changed to Closed",
+                                    ),
+                                ))?;
+                            },
```
ThetaSinner: I question these being handled as errors. I see that this matches the behavior in the data channel implementation but it will lead to a `tracing::error!` with a "WebRTC task failed" message. That seems misleading to me. Should these be their own enum variant to communicate stopping without an error?

Possibly something that should be considered in the data channel implementation too.

mattyg: That makes sense to me that this is proper handling of expected behavior rather than an error.

I would suggest doing that refactoring in a separate PR because I'm guessing it will be a wider scope. I believe we're relying on the errors propagating up to trigger dropping the connection, but I haven't combed through the codebase to see it clearly.

crates/tx5-connection/src/webrtc/go_pion.rs
mattyg: I'm don't think either approach (including mine) really asserts anything reliably -- it takes some unknown time for the peer connection state change to be handled, so with either approach we're still just asserting "the connection is still open at this time" -- but it could still close after which would not cause the assertion to fail. I don't know a correct way to test this.

jost-s (commented):
Reviewed this and looks good to me

crates/tx5-connection/src/webrtc/go_pion.rs
```diff
@@ -107,15 +108,37 @@ async fn task(
...
+    task_inner(
+        is_polite,
+        send_buffer,
+        evt_send,
+        cmd_send,
+        cmd_recv,
+        peer,
+        peer_evt,
+    )
+    .await
```
neonphog: I'm not following the point of this outer function. Usually I do this because I need to capture the error and do something with it. In this case, couldn't the two lines above just move down into the inner function?

neonphog (approved):
This looks fine to me. The test isn't really testing much... just that the event gets re-published. But it's probably the best you can do for a unit test. It'd be good to add some integration tests of this at some point.

crates/tx5-connection/src/webrtc/go_pion.rs
ThetaSinner: The errors are staying in this file and just being logged. What we rely on is the drop behavior. I don't think the scope of that change would be any more significant than what you've already done by adding the `Error` variant. On another PR is fine but then can we track it please :)

crates/tx5-connection/src/webrtc/go_pion.rs
```diff
@@ -124,8 +148,25 @@ async fn task(
...
+                    Evt::State(state) => match state {
+                        tx5_go_pion::PeerConnectionState::Disconnected => {
+                            cmd_send2.send_or_close(Cmd::Close)?;
+                        }
+                        tx5_go_pion::PeerConnectionState::Closed => {
+                            cmd_send2.send_or_close(Cmd::Close)?;
+                        }
+                        tx5_go_pion::PeerConnectionState::Failed => {
+                            cmd_send2.send_or_close(Cmd::Close)?;
+                        }
```
ThetaSinner: I was just after changing the `Disconnected` and `Closed` events, is this one not an error? It's not very clear from their docs but I'd expect it to be

ThetaSinner (commented):
Build seems to be broken, and why is `ConnCmd` also now changing? Is that required because of changing events in the webrtc module?

crates/tx5-connection/src/webrtc/go_pion.rs
mattyg: Yeah I agree its not very clear from the spec. Let's go with error https://github.com/holochain/tx5/pull/194/commits/e84fceae60304513ecf80c40a7ad3a99cba61878

crates/tx5-connection/src/webrtc/go_pion.rs
mattyg: Deleted the assertion

crates/tx5-connection/src/webrtc/go_pion.rs
mattyg: I'm can't figure out how to do a proper integration test that exercises this. Open to any suggestions.

ThetaSinner (dismissed):

ThetaSinner (approved):

PR #199, opened 2025-11-11
fix: don't send Disconnect until peer has been removed from peer_map

mattyg:
This is 2 changes to reduce test flakes of `sending_second_message_after_remote_disconnect_succeeds_after_disconnect_event`.

This PR seems to eliminate the flakes completely. On `main` it consistently failed with 10 test runs. In this PR it does not arise within 500 test runs with either backend. The test run that reproduced it was `sending_second_message_after_remote_disconnect_succeeds_after_disconnect_event`.

The changes are explained in inline comments.

## Summary by CodeRabbit

* **Bug Fixes**
  * Peer disconnects now use a 10-second timeout and log failures/timeouts to improve reliability and diagnostics.
  * Cleanup of peer connections is now driven by an explicit guard to avoid premature or immediate disconnects, reducing race conditions during shutdown and tests.

### DIFF
--- a/crates/tx5/src/peer.rs
+++ b/crates/tx5/src/peer.rs
@@ -189,10 +189,31 @@ impl Drop for DropPeer {
 
         let evt_send = self.evt_send.clone();
         let peer_url = self.peer_url.clone();
+
+        // Notify the endpoint of disconnect
         tokio::task::spawn(async move {
-            let _ = evt_send
-                .send(EndpointEvent::Disconnected { peer_url })
-                .await;
+            match tokio::time::timeout(
+                std::time::Duration::from_secs(10),
+                evt_send.send(EndpointEvent::Disconnected {
+                    peer_url: peer_url.clone(),
+                }),
+            )
+            .await
+            {
+                Ok(Ok(())) => {}
+                Ok(Err(_)) => {
+                    tracing::debug!(
+                        ?peer_url,
+                        "failed to send disconnected event"
+                    );
+                }
+                Err(_) => {
+                    tracing::debug!(
+                        ?peer_url,
+                        "timed out sending disconnected event"
+                    );
+                }
+            }
         });
     }
 }
@@ -219,7 +240,7 @@ async fn task(
     ready: MaybeReady,
 ) {
     // establish our cleanup drop guard
-    let _drop = DropPeer {
+    let drop_guard = DropPeer {
         ep,
         peer_url: peer_url.clone(),
         evt_send: evt_send.clone(),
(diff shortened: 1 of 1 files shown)


crates/tx5-connection/src/hub.rs
```diff
@@ -239,14 +239,15 @@ impl Hub {
                         }
                     }
                     HubCmd::Disconnect(pub_key) => {
-                        // disconnect from a remote peer
+                        // Remove remote peer from hub map
```
mattyg: First remove the peer from the hub map, then call close_peer.

If `DynBackEp::connect` is called before this `close_peer` future resolves, then that connection attempt would fail with an error "already connected".

Since we are ignoring the result from `close_peer` here anyway, this does not change the logic.

crates/tx5/src/peer.rs
```diff
@@ -339,23 +360,5 @@ async fn task(
             .await;
     }
 
-    // the above loop ended, notify of disconnect
-    match tokio::time::timeout(
```
mattyg: By sending the disconnect here, it is sent before the peer is removed from the peer map.

Thus we may remove close the sbd connection for the peer and remove it from the hub map, but still have a Peer available that can be interacted with. Those interactions will fail.

Instead this change is to only send the disconnect when the Peer is dropped.

crates/tx5-connection/src/conn.rs
```diff
@@ -238,12 +238,28 @@ impl TaskCore {
...
 }
+/// Drop guard that closes the ready semaphore. This ensures callers will not be waiting
+/// for ready indefinitely after the task has ended.
+struct DropReady {
+    ready: Arc<tokio::sync::Semaphore>,
+}
+
+impl Drop for DropReady {
+    fn drop(&mut self) {
+        self.ready.close();
```
ThetaSinner: I don't think it's acceptable to close this if a ready state is not reached. Closing the ready semaphore allows the code to proceed as though the connection can be used and we then rely on something downstream of the ready check failing and allowing everything to shut down. If nothing else, that is hard to reason about when reading the code.

The failure to ready is already handled in https://github.com/holochain/tx5/blob/main/crates/tx5/src/backend/be_tx5_connection.rs#L58 which will catch the case that something fails. It's not ideal and not as sophisticated as the solution higher up the stack. Changing that though, I think should look like communicating success or failure with the semaphore close, and not just closing it and letting the code proceed as though the connection is ready.

crates/tx5/src/peer.rs
```diff
@@ -339,23 +360,10 @@ async fn task(
...
-        }
-    }
-
-    // all other cleanup is handled by the drop guard
+    // Drop the drop guard to run cleanup tasks.
+    //
+    // If the Peer is dropped early, then the DropPeer guard will be dropped with it.
+    // Otherwise if the above loop ends before the Peer is dropped,
+    // we drop the DropPeer guard manually to run cleanup tasks.
+    drop(drop_guard);
```
ThetaSinner, line 373: Should not be necessary to drop explicitly here, but the expanded comment is useful

ThetaSinner (changes requested):
I think the change to peer.rs makes sense, I don't think the change to conn.rs does. Does one work without the other? Dealing with the ready success/failure states is a bigger change

crates/tx5/src/peer.rs
mattyg: I agree, but without this a test does consistently fail within ~250 runs. With this it doesn't. I can't explain why, but could dig deeper to find out. Or we could keep in and I could note that in a comment. Or we could remove it and not worry about the edge case failure.

neonphog: I also can't imagine why an explicit drop helps, but I totally believe you. Yes, I'd say add a comment to that effect and leave this in.

crates/tx5-connection/src/conn.rs
neonphog: Indeed, making tasks wait a timeout when we know already it is not going to work isn't ideal...

It may be worth a quick spike to make the ready function (line 153) return a result to see how complicated an update that is. If it's not bad, the ready semaphore could be changed to a tokio oneshot that returns a `Result<()>`.

To start, the drop case could return a disconnect error, and later on we could send back more detailed errors for the connection failure branches.

neonphog (changes requested):
I think it's worth spending a little time to attempt making the ready return more semantic results. If that turns out to be a rabbit hole, I'm pretty comfortable with these changes, maybe just adding a comment that closing the ready without somehow better indicating it is an error case is not ideal.

mattyg:
For the sake of not putting too much time into tx5 edge case issues, I'm going to keep in the `peer.rs` change with an explanation comment and remove the `conn.rs` change rather than do additional refactoring.

This significantly reduces test flakes such that it shouldn't disrupt our productivity, so good enough to move forward.

ThetaSinner (approved):

PR #200, opened 2025-11-11
fix: deadlock from not closing Conn ready semaphore

mattyg:


mattyg:
Closing as this logically doesn't appear to do anything and the issue was caused by another fix PR that is a rare enough edge case that its not worth including.

